PR #2, opened 2018-04-12
Updates to Atlas CLI

ZachEddy:
# Updates to Atlas CLI

This pull request fixes a few major issues that I encountered when helping teammates bootstrap new applications.

- **Flexible Go Environment** — The Atlas CLI now initializes new projects no matter how the developer has defined his or her `GOPATH` environment variable. With the changes I've added, both of the following `GOPATH` examples work: `/Users/person/go` or `/work/go_projects/`

- **Fix Incomplete Imports** — Applications files are no longer bootstrapped with broken or incomplete Go imports.

- **Helpful Error Messages** — The `init-app` command executes several subcommands as part of the boostrapping process (e.g. `dep init` to initialize the `Gopkg.toml` and `Gopkg.lock` files). If these subcommands fail, the Atlas CLI will now print helpful error messages.

- **README Change** — I replaced a reference to `ngp.api.toolkit` with `atlas.app.toolkit`


cli/atlas/formatting.go
```diff
@@ -40,17 +41,23 @@ func ServerURL(str string) (string, error) {
...
-	dirs := strings.Split(dirString, "/")
-	for i, dir := range dirs {
-		if strings.ToLower(dir) == "go" && i+1 < len(dirs) {
-			if i+2 < len(dirs) && strings.ToLower(dirs[i+1]) == "src" {
-				return strings.Join(dirs[i+2:], "/"), nil
-			}
-		}
+// root is considered to be anything in $GOPATH/src or below
+func ProjectRoot(gopath, workdir string) (string, error) {
+	if gopath == "" {
```
dmacthedestroyer: technically, if no gopath is provided you could get it yourself: https://stackoverflow.com/a/32650077/437456

cli/atlas/formatting_test.go
```diff
@@ -57,38 +57,68 @@ func TestServerURL(t *testing.T) {
...
+			workdir:  "go/src/github.com/SomeProject",
+			expected: "github.com/SomeProject",
+			err:      nil,
 		},
 		{
-			"go/src",
-			"",
-			errInvalidProjectRoot,
+			gopath:   "/Users/person/go",
+			workdir:  "/Users/person/go/src/github.com/infobloxopen/SomeProject",
```
dmacthedestroyer: our project naming conventions is to use kebab-case ;)

cli/atlas/formatting.go
```diff
@@ -40,17 +41,23 @@ func ServerURL(str string) (string, error) {
...
+		return "", errMissingGOPATH
 	}
-	return "", errInvalidProjectRoot
+	if len(workdir) < len(gopath) {
```
dmacthedestroyer: do you use the `gopath` for anything other than this validation? I'm wondering why we care so much if it's not valid.

A valid use case could be that someone bootstraps their app in some dumb file location then after that change their gopath to point to it. They'd be dumb, but who are we to judge?

An alternative would be to give them a warning if their location isn't in their gopath, instead of a hard fail.

cli/atlas/main.go
```diff
@@ -154,7 +155,7 @@ func initializeApplication() {
 	if err != nil {
 		printErr(err)
 	}
-	root, err := ProjectRoot(wd)
+	root, err := ProjectRoot(os.Getenv("GOPATH"), wd)
```
dmacthedestroyer: oh, I see -- you are getting the gopath on your own.  Nice work! :)

cli/atlas/formatting.go
dmacthedestroyer: I'm relatively unopinionated on it, just providing an alternative perspective. I'll be happy with whichever choice you think is most useful :)

cli/atlas/main.go
```diff
@@ -227,11 +226,24 @@ func generateProtobuf() error {
...
-		if err != nil {
+		if err := runCommand("goimports", "-w", dir); err != nil {
 			return err
 		}
 	}
 	fmt.Println("done!")
 	return nil
 }
+
+func runCommand(command string, args ...string) error {
```
dmacthedestroyer: very nice - good work!

cli/atlas/main.go
```diff
@@ -227,11 +226,24 @@ func generateProtobuf() error {
...
+	cmd := exec.Command(command, args...)
+	stderr := &bytes.Buffer{}
+	cmd.Stderr = stderr
```
dmacthedestroyer: feature request: capture `stdout` as well and print it if a -debug flag is present :)

ZachEddy: Nice idea! I'll add that.

cli/atlas/formatting.go
ZachEddy: IMO, people should have to initialize projects in their `GOPATH` because it's more in line with convential Go practices.

PR #4, opened 2018-04-16
Update gRPC version from v1.10.0 to v1.11.0

ZachEddy:
## Update gRPC Version

This pull request moves the toolkit from `grpc-go` v1.10.0 to `grpc-go` 1.11.0 due to a recent change that broke backward compatibility.

I also added unit tests for `WithRequest()` because the newest version of `grpc-go` now enables users to create a mock `transport.Stream` (see [this](https://github.com/grpc/grpc-go/pull/1904) for more information).

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -61,7 +61,7 @@
 
 [[constraint]]
   name = "google.golang.org/grpc"
-  version = "1.10.0"
+  version = "1.11.0"
 
 [prune]
   go-tests = true
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -217,12 +217,12 @@
     "tap",
     "transport"
   ]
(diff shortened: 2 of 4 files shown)


Gopkg.toml
```diff
@@ -61,7 +61,7 @@
 
 [[constraint]]
   name = "google.golang.org/grpc"
-  version = "1.10.0"
+  version = "1.11.0"
```
dmacthedestroyer: do we need to up the dependency here? We're no longer supporting v1.10.0?

ZachEddy:
@dmacthedestroyer Unfortunately, we have to update the dependency here. The changes introduced in version v1.11.0 are backward-incompatible, so anything below v1.11.0 will break the authorization code.

Because application teams haven't adopted the toolkit yet, it's safe to bump the version. We'll have to tread carefully around version changes once teams start using the toolkit.

PR #5, opened 2018-04-16
Update JWT Handling

ZachEddy:
## Update JWT Handling

This pull request mostly reorganizes authorization code. Specifically, I've moved all of the jwt-related code into the `jwt.go` and `jwt_test.go` files.

As far as new code goes: I created a function called `GetJWTField` that lets users get specific fields from the JWT payload. The `GetAccountID` (formerly `GetTenantID`) function uses this function directly.


mw/auth/jwt_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	jwt "github.com/dgrijalva/jwt-go"
+	"google.golang.org/grpc/metadata"
+)
+
+const (
+	TEST_SECRET = "some-secret-123"
+)
+
+func TestGetTenantID(t *testing.T) {
```
dmacthedestroyer: bad name!

mw/auth/jwt_test.go
```diff
@@ -0,0 +1,83 @@
...
+	var tenantIDTests = []struct {
```
dmacthedestroyer: bad name!

mw/auth/jwt_test.go
```diff
@@ -0,0 +1,83 @@
...
+			err:      errMissingField,
+		},
+		{
+			claims:   jwt.MapClaims{},
+			field:    "some-field",
+			expected: "",
+			err:      errMissingToken,
+		},
+	}
+	for _, test := range tenantIDTests {
```
dmacthedestroyer: this doesn't seem to test the `GetAccountID` function anymore at all, instead testing the `GetJWTField` function.

You should also have tests for the `GetAccountID` function -- it can be a small, easy test.

mw/auth/jwt_test.go
```diff
@@ -0,0 +1,83 @@
...
+		ctx := context.Background()
+		if len(test.claims) != 0 {
+			token := makeToken(test.claims, t)
+			ctx, _ = contextWithToken(token)
```
dmacthedestroyer: is the `_` an `error`? If so, you should check it and call a `t.Fatalf` if it exists.

ZachEddy: Yeah, that returns an error. Sloppy coder strikes again.

mw/auth/jwt_test.go
ZachEddy: Makes sense!

mw/auth/jwt_test.go
ZachEddy: Another nice catch!

mw/auth/jwt_test.go
ZachEddy: Nice catch! That's some sloppy coding.

PR #6, opened 2018-04-17
WIP: Add common server packages and options

johnbelamaric:
I noticed that the server generated by the `atlas` command line tool does not do TLS. We need to do TLS, and in fact we need to do it in a manner that enforces client certificates that have been signed by our CA (and maybe more stringent later).

The server also does not expose health or metrics endpoints. 

So, I started putting together the framework for how I would like to see this work. We need server struct that manages the grpc, health, and metrics endpoints. This server needs to be able to be initialized with functional options that enable the various pieces. For example, in dev, we may not want to bother with TLS, so we need command line flags to modify that behavior.

The result is in `pkg/server`. Right now, it handles gRPC with TLS transport credentials. There are some additional things started in `pkg/flag` to enable health and metrics endpoints. I am sure we will come up with more uses for it, too.

### DIFF
--- a/pkg/flag/flag.go
+++ b/pkg/flag/flag.go
@@ -1,0 +1,67 @@
+package flag
+
+import (
+	"crypto/tls"
+	"flag"
+
+	atlastls "github.com/infobloxopen/atlas-app-toolkit/pkg/tls"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+type GRPCFlags struct {
+	addr *string
+}
+
+type TLSFlags struct {
+	cert *string
+	key *string
+	ca *string
+}
+
+func HealthFlags() (*string, *string) {
+	addr := flag.String("health-addr", ":8080", "address to use for the health endpoint")
+	path := flag.String("health-path", "/health", "path to use for the health endpoint")
+	return addr, path
+}
+
+func MetricsFlags() (*string, *string) {
+	addr := flag.String("metrics-addr", ":9500", "address to use for the metrics endpoint")
+	path := flag.String("metrics-path", "/metrics", "path to use for the metrics endpoint")
+	return addr, path
+}
+
+func NewGRPCFlags() *GRPCFlags {
+	f := &GRPCFlags{}
+	f.addr = flag.String("addr", ":9000", "address to use for the gRPC endpoint")
+	return f
+}
+
+func (f *GRPCFlags) Addr() string {
+	return *f.addr
+}
+
+func NewTLSFlags() *TLSFlags {
+	f := &TLSFlags{}
+	f.cert = flag.String("cert", "", "path to the server certificate in PEM format")
+	f.key = flag.String("key", "", "path to the server private key in PEM format")
+	f.ca = flag.String("ca", "", "path to the CA PEM for validating the client cert; system CAs will be used if blank")
+
+	return f
+}
+
+func (f *TLSFlags) TLSConfig() (*tls.Config, error) {
+	return atlastls.NewTLSConfig(*f.cert, *f.key, *f.ca)
+}
+
+func (f *TLSFlags) WithGRPCTLSCreds() (grpc.ServerOption, error) {
+	if *f.cert == "" {
+		return nil, nil
+	}
+	tlsConfig, err := f.TLSConfig()
+	if err != nil {
+		return nil, err
+	}
+	return grpc.Creds(credentials.NewTLS(tlsConfig)), nil
+}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1,0 +1,130 @@
+package server
+
+import (
+	"crypto/tls"
+	"net"
+	"net/http"
+
+	atlastls "github.com/infobloxopen/atlas-app-toolkit/pkg/tls"
+
+	"google.golang.org/grpc"
+)
+
+type Server struct {
+	grpcAddr    string
+	healthAddr  string
+	healthPath  string
+	metricsAddr string
+	metricsPath string
+	tlsConfig   *tls.Config
+	grpcServer  *grpc.Server
+	httpMux     *http.ServeMux
+}
+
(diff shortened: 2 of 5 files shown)


johnbelamaric:
Oh, and here is what a main file might look like (based on the `cmd/service/main.go`):

```
package main

import (
	"flag"

	atlasflag "github.com/infobloxopen/atlas-app-toolkit/pkg/flag"
	"github.com/infobloxopen/atlas-app-toolkit/pkg/server"

	"github.com/grpc-ecosystem/go-grpc-middleware"
	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus"
	"github.com/grpc-ecosystem/go-grpc-middleware/validator"
	"github.com/infobloxopen/sample/pb"
	"github.com/infobloxopen/sample/svc"
	"github.com/sirupsen/logrus"
	"google.golang.org/grpc"
)

var (
	grpcFlags *atlasflag.GRPCFlags
	tlsFlags *atlasflag.TLSFlags
)

func main() {
	logger := logrus.New()

	opts := append([]grpc.ServerOption{},
		grpc.UnaryInterceptor(
			grpc_middleware.ChainUnaryServer(
				// validation middleware
				grpc_validator.UnaryServerInterceptor(),
				// logging middleware
				grpc_logrus.UnaryServerInterceptor(logrus.NewEntry(logger)),
			),
		),
	)

	tlsOpt, err := tlsFlags.WithGRPCTLSCreds()
	if err != nil {
		logger.Fatalln(err)
	}

	if tlsOpt != nil {
		opts = append(opts, tlsOpt)
	}

	// create new gRPC server with middleware chain
	grpcServer := grpc.NewServer(opts...)

	// register service implementation with the grpc server
	svcServer, err := svc.NewBasicServer()
	if err != nil {
		logger.Fatalln(err)
	}
	pb.RegisterSampleServer(grpcServer, svcServer)

	s, err := server.New(
		server.WithGRPC(grpcFlags.Addr(), grpcServer),
	)

	if err != nil {
		logger.Fatalln(err)
	}

	if err := s.Serve(); err != nil {
		logger.Fatalln(err)
	}
}

func init() {
	grpcFlags = atlasflag.NewGRPCFlags()
	tlsFlags = atlasflag.NewTLSFlags()
	flag.Parse()
}
```

johnbelamaric:
@ZachEddy @dmacthedestroyer @Calebjh PTAL

johnbelamaric:
Looks like the code that launches health endpoint is currently missing here (same for metrics below).

Correct, this is just showing the basic framework, it’s not complete.

Should health be started before grpc server? What if grpc server is waiting for db connection? This may causethat pod will be restarted. I would suggest to start health and ready endpoints first and once grpc server is started readiness will change from false to true.

That makes sense. I was thinking we didn’t want to return healthy until we at least started grpc but it doesn’t actually matter - and readiness check makes more sense there. In practice it doesn’t make a lot of difference, because these are all started in go routines so should all come up within nanoseconds of each other.

For health and readiness we need to provide a way to register callbacks, like WithHealthCallback() that takes a function or WithHealther() that takes an interface. Same for readiness. Then developers can influence health/readiness themselves. We probably want a ”WithGorm()” … maybe … that might come just directly from the gRPC middleware. Not clear to me yet. Also, does gorm support connection pooling?

johnbelamaric:
*   It is not clear for me why we have "pkg" in path for all new files. Can we think of a better naming structure.

That’s standard for organizing go code reusable bits.

  *   Current code has only health, but not readiness. Should we have separate WithHealth() & WithReadiness() or somehow combine this? Also we already have health/readiness support which will need to be updated accordingly if we will follow given pattern.

Agreed with need WithReadiness() too; but I think it can use the same HTTP endpoint. For that matter so could metrics.

pkg/flag/flag.go
```diff
@@ -0,0 +1,67 @@
...
+	f.ca = flag.String("ca", "", "path to the CA PEM for validating the client cert; system CAs will be used if blank")
+
+	return f
+}
+
+func (f *TLSFlags) TLSConfig() (*tls.Config, error) {
+	return atlastls.NewTLSConfig(*f.cert, *f.key, *f.ca)
+}
+
+func (f *TLSFlags) WithGRPCTLSCreds() (grpc.ServerOption, error) {
```
vitalykarpenko: If we stick to this design, the resulting code from user side will be something like:
``` golang
tlsFlags := NewTLSFlags()
optWithTLS, err := tlsFlags.WithGRPCTLSCreds()
if err != nil {
    panic(err)
}
srv := grpc.NewServer(optWithTLS)
```
This leads us to a predefined set of flags for accepting TLS certs and keys. I see no reason to move TLS config initialization into toolkit. It can be googled in 5 seconds and implemented using grpc only code without any restrictions on keys/certs falg names/paths or anything else:
``` golang
credentials.NewServerTLSFromFile("cert/server.crt", "cert/server.key")
if err != nil {
    log.Fatalf("could not load TLS keys: %s", err)
}
grpcServer := grpc.NewServer(grpc.Creds(creds))
```

johnbelamaric: We want a consistent interface to all apps built with the toolkit. That includes flags - there is another story being worked to allow consistent configuration via flags, env vars, and file. Without that, you are hard-coding the paths to the certs, which limits deployment flexibility substantially.

vitalykarpenko: It's not about hard-coding paths to certificates. Our flags(for tls config) currently limiting user from having any alternative type of input of tls-related things. Moreover, we want him to use our code for creating grpc.ServerOption for his grpc server which can be created only by him and only inside his code with any grpc.ServerOptions he wants.

johnbelamaric: We want the consistent flags. That specific flag mechanism needs to be expanded into the general config mechanism that uses any of flag/env var/file. That should move to some other pkg for that purpose.

Nothing actually forces the author to use this - they can still do it exactly how you did above, if they want different flags or whatever for some reason. It's just a convenience.

PR #22, opened 2018-04-28
Added ServerWrapper structure

vitalykarpenko:
Introduced health checks addition as an option


johnbelamaric:
Ok, good start. Don't really like the name though. Maybe something like "AppServer" or something? I'd rather not instantiate a "Wrapper" in the main.go.

We really need the TLS stuff from #6 too.

ZachEddy:
Go kit uses [a similar design pattern](https://github.com/go-kit/kit/blob/master/transport/grpc/server.go) for creating new gRPC servers. It might have some design pearls that we can bring to the toolkit, so I wanted to share. It also uses this pattern for more RESTful approaches, which makes me think we can export `NewGateway` and `NewServer` functions. Just some food for thought.

pkg/appserver/appserver.go
```diff
@@ -0,0 +1,167 @@
...
+			if err := s.grpcServer.Serve(g); err != nil {
+				errChan <- err
+				return
+			}
+			doneChan <- true
+			return
+		}()
+	}
+	if s.metricsAddr != "" && m != nil {
+	}
```
dmacthedestroyer: I'm guessing the idea here is that this will call a goroutine that runs a listener, much like the health check?

dmacthedestroyer:
@vitalykarpenko are you still working on this? I really like the pattern and would like to see it in master

Evgeniy-L:
@dmacthedestroyer, Vitaly is out for 3 weeks, we should figure out what to do with this PR

PR #25, opened 2018-05-02
Create new gRPC gateways with app toolkit

ZachEddy:
# REST Gateway Functional Options

Currently, spinning-up a new REST gateway isn't straightforward. It would be nice to provide developers with some gateway-specific options that abstract away boilerplate code and toolkit-specific configuration. That's what this pull request attempts to introduce.

IMO, here are some of the key advantages:

- Less mental overhead for developers because gateway-related nuances are shifted to the toolkit code.
- It avoids repeating the same code from application-to-application.
- Fewer lines of code in the `main.go` file, which is especially helpful because we've decided to package the gRPC server and REST gateway in the same binary.

As a final note: it seems weird to have a `gw` and `pkg/gw` or `pkg/gateway` directory. If we decide to have a `pkg/` directory, it might be worth restructuring the toolkit and cutting a new release. Also, I'd like to add tests for this, but I'm not quite sure how to do so.


ZachEddy:
[Here's](https://gist.github.com/ZachEddy/04b9e5f5607fc468a44854874574c28e) an example usage of the gateway functional options.

pkg/gateway/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+	serverDialOptions []grpc.DialOption
+	endpoints         map[string][]registerFunc
+}
+
+// NewGateway creates a gRPC REST gateway with HTTP handlers that have been
+// generated by the gRPC gateway protoc plugin
+func NewGateway(options ...Option) (*http.ServeMux, error) {
+	// configure gateway defaults
+	g := gateway{
+		serverAddress:     ":9090",
```
dmacthedestroyer: it would be great to expose this as an exported `const` instead of a string literal

ZachEddy:
@johnbelamaric What are your thoughts on this approach?

johnbelamaric:
👍 nice!

PR #28, opened 2018-05-04
Add -db Flag to Atlas CLI

ZachEddy:
# Add -db Flag to Atlas CLI

This pull requests introduces the `-db` flag. When the `-db` flag is provided to the boostrap command, here's what will happen:
- The `protoc-gen-gorm` plugin is included in the `protoc` command (inside the Makefile).
- The default server implementation is generated with a `gorm.DB` field.
- The `config.go` file will include a default database address.

[Here's](https://github.com/ZachEddy/atlas-app-toolkit/tree/cli-add-db-flag-example/cli/example) an example project that was generated with the `-db` flag.


cli/atlas/templates/cmd/config.go.gotmpl
```diff
@@ -8,5 +8,7 @@ const (
   GatewayAddress = "0.0.0.0:8080"
   // GatewayURL is the default URL endpoint for the gateway server.
   GatewayURL = "/{{ .Name | URL }}/v1/"
-  {{ end }}
+  {{ end }}{{ if .WithDatabase -}}
+  // DatabaseAddress is the default address for the database, if no override is specified in the flags
+  DatabaseAddress = "0.0.0.0:5432"{{ end }}
```
dmacthedestroyer: this should be `DatabaseConnectionString` or `DBConnectionString` instead of address, and default to a postgres connection string to a database named after the generated `-name` for the project:
`"host=localhost port=5432 dbname=<app name>"`

There might be some TLS thing you'll need in there, too...

And this should probably also include running a postgres docker image with the same configuration in the Makefile

cli/atlas/templates/cmd/server/main.go.gotmpl
```diff
@@ -9,10 +9,12 @@ import (
 	"github.com/grpc-ecosystem/go-grpc-middleware/validator"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
+	{{ if .WithDatabase }}"github.com/jinzhu/gorm"{{ end }}
 )
 
 var (
 	Address string
+	Database string
```
dmacthedestroyer: change to `DBConnectionString` (or something)

cli/atlas/templates/pkg/pb/service.proto.gotmpl
```diff
@@ -5,7 +5,8 @@ package service;
 import "google/protobuf/empty.proto";
 import "google/api/annotations.proto";
 import "github.com/lyft/protoc-gen-validate/validate/validate.proto";
-import "protoc-gen-swagger/options/annotations.proto";
+import "protoc-gen-swagger/options/annotations.proto";{{ if .WithDatabase }}
```
dmacthedestroyer: nitpick: I feel like these templates would all be cleaner if the whole conditional statement was on one line... why does it need to be split across two lines like this?

cli/atlas/templates/pkg/svc/zserver_test.go.gotmpl
```diff
@@ -20,7 +20,7 @@ func TestGetVersion(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			server, err := NewBasicServer()
+			server, err := NewBasicServer({{ if .WithDatabase }}nil{{ end }})
```
dmacthedestroyer: hmm, this provides a good opportunity to demonstrate how to properly stub a database for a unit test.

Or maybe we should set this up to be an integration test, with the whole `e2e` flag and whatnot?

dmacthedestroyer: yeah! something kinda like that. I don't have a really clear answer, unfortunately, and I don't think anyone else here in Tacoma has an example of one :(

cli/atlas/templates/cmd/config.go.gotmpl
dmacthedestroyer: *The name thing:* Good point -- it could be weird. Maybe spend a few seconds to see what kind of database naming restrictions there are for postgres? I think we should keep to just the `-name` parameter and let it look weird rather than adding a new name -- that can be a problem for Future Zach to worry about :)

*The TLS thing:* I meant there should be a `sslmode=disable` setting in the connection string, like mentioned [here](https://github.com/xo/usql/issues/6). I don't know if it's necessary, but it was for me out-of-the-box. It could have been that I did something dumb, though :)

*postgres docker thing:* yep, ignore this in favor of getting a bootstrapped manifest. Maybe that can be worked on really soon to obviate this need.

Issue #24, opened 2018-04-30
Collection operator middleware should match gRPC-gateway convention


dmacthedestroyer:
Currently, you reference the collection operators middleware like this:
```
mw.WithCollectionOperator()
```
We should instead change `WithCollectionOperator` to match [grpc-ecosystem](https://github.com/grpc-ecosystem/go-grpc-middleware#middleware)'s naming convention:
```
collectionoperators. UnaryServerInterceptor
```
or something like that.

Calebjh:
There are other middlewares now that break this naming convention (although less egregiously for the most part, I think, generally including "Interceptor" somewhere in the name), but this middleware has since been renamed and this issue is outdated.

PR #29, opened 2018-05-08
Rename collection operator middleware

kumaya:
- refactor the name to match gRPC-gateway convention
  - fix minor issue in README.md
  - fix test cases to match new naming convention
  - referencing issue #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ For this purpose `auth.GetAccountID(ctx, nil)` function can be used:
 func (s *contactsServer) Read(ctx context.Context, req *ReadRequest) (*ReadResponse, error) {
 	input := req.GetContact()
 
-	accountID, err := mw.GetAccountID(ctx, nil)
+	accountID, err := auth.GetAccountID(ctx, nil)
 	if err == nil {
 		input.AccountId = accountID
 	} else if input.GetAccountId() == "" {
@@ -575,7 +575,7 @@ message MyRequest {
 ```
 
 After you declare one of collection operator in your `proto` message you need
-to add `mw.WithCollectionOperator` server interceptor to the chain in your
+to add `collection_operators.UnaryServerInterceptor` server interceptor to the chain in your
 gRPC service.
 
 ```golang
@@ -583,7 +583,7 @@ gRPC service.
     grpc.UnaryInterceptor(
       grpc_middleware.ChainUnaryServer( // middleware chain
         ...
-        mw.WithCollectionOperator(), // collection operators
+        collection_operators.UnaryServerInterceptor(), // collection operators
         ...
       ),
     ),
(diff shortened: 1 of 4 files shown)


ZachEddy:
@dmacthedestroyer To clarify: is this what you have in mind?

```sh
# atlas-app-toolkit/middleware
$ tree -A -C middleware
middleware
├── README.md
├── auth
│   ├── interceptor.go
│   ├── interceptor_test.go
│   ├── jwt.go
│   ├── jwt_test.go
│   ├── options.go
│   └── options_test.go
└── collections
    ├── operator.go
    └── operator_test.go
```
In this case, `collections` will be the new package name.
```sh
# atlas-app-toolkit/operators
$ tree -A -C operators
operators
├── collection_operators.pb.go
├── collection_operators.proto
├── fields.go
├── fields_test.go
├── filtering.go
├── filtering_lexer.go
├── filtering_lexer_test.go
├── filtering_parser.go
├── filtering_parser_test.go
├── filtering_test.go
├── gorm
│   ├── collection_operators.go
│   ├── collection_operators_test.go
│   ├── filtering.go
│   └── filtering_test.go
├── pagination.go
├── pagination_test.go
├── sorting.go
└── sorting_test.go
```

dmacthedestroyer:
@ZachEddy Hmm... maybe it should be `collections` at both packages? I kinda sucks having the same name for them, but it also kinda sucks calling one `operators` and the other `collections` when they're the same thing.

The other option is that we don't have a `middleware` package and instead put the middlewares in their respective packages. In this case there'd be a `collections/middleware` package and an `auth/middleware` package. I'm kind of leaning towards that, now.

ZachEddy:
> The other option is that we don't have a middleware package and instead put the middlewares in their respective packages. In this case, there'd be a collections/middleware package and an auth/middleware package. I'm kind of leaning towards that, now.

I'm a fan of this approach. Do you think we can remove the `/middleware` subdirectory altogether? As in, we'll just have `atlas-app-toolkit/auth` and `atla-app-toolkit/collections`? I'm trying to think of the most developer-friendly approach.

dmacthedestroyer:
> Do you think we can remove the /middleware subdirectory altogether?
I'm a fan of this approach :)

In this case, it'd be `collections.UnaryServerInterceptor()` and `auth.UnaryServerInterceptor()` or something like that. Right?

ZachEddy:
Yep! In its current state, authorization middleware is added by calling `auth.DefaultAuthInterceptor()`, so that will need to get changed.

kumaya:
Removing mw and moving operators to collection (renamed op) impose issue of cyclic imports, where mc/operators.go import gw and vice versa gw/fields.go imports op.

dmacthedestroyer:
what a mess! You [nerd sniped](https://xkcd.com/356/) me and I took a stab at this with PR #31. Take a look at it and let me know what you think!

dmacthedestroyer:
closing in favor of #31

PR #32, opened 2018-05-10
Good to include read and write timeouts in server

chinmayb:
Its to enforce timeouts on client connections.

`ReadTimeout` covers the time from when the connection is accepted to when the request body is fully read.
`WriteTimeout` covers the time from the end of the request header read to the end of the response.


dmacthedestroyer:
@ZachEddy any changes you'd like to see from this PR? If not, feel free to merge it

PR #33, opened 2018-05-11
Update WithRequest option to include an appID argument

ZachEddy:
# Add appID Argument to WithRequest

The current iteration of the `WithRequest` option isn't quite what we want. It uses the microservice name (e.g. PetStore) as the application name. To resolve this issue, developers will now provide an application ID (e.g. ShoppingMall) to the `WithRequest` option. The application ID will associate their service to a specific application.

Operations in PARG files should now be prefixed with a service name. Take a look at [this pull request](https://github.com/infobloxopen/atlas-contacts-app/pull/17) to see up-to-date PARGs.

_This is a breaking change_

### DIFF
--- a/auth/interceptor.go
+++ b/auth/interceptor.go
@@ -143,10 +143,10 @@ func makeInterceptor(a string, b Builder, h Handler) grpc.UnaryServerInterceptor
 	return grpc_auth.UnaryServerInterceptor(authorizer.AuthFunc())
 }
 
-func UnaryServerInterceptor(authzAddress string) grpc.UnaryServerInterceptor {
+func UnaryServerInterceptor(authzAddress, appID string) grpc.UnaryServerInterceptor {
 	return makeInterceptor(
 		authzAddress,
-		NewBuilder(WithJWT(nil), WithRequest()),
+		NewBuilder(WithJWT(nil), WithRequest(appID)),
 		NewHandler(),
 	)
 }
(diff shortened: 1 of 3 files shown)


mw/auth/options.go
```diff
@@ -55,17 +55,21 @@ func WithCallback(attr attributer) option {
 // to Themis in the authorization request. Specifically, this includes the gRPC
 // service name (e.g. AddressBook) and the corresponding function that is
 // called by the client (e.g. ListPersons)
-func WithRequest() option {
+func WithRequest(appID string) option {
+	// assume PARGs are in default namespace if no appID is provided
```
johnbelamaric: @markinos - are we still equating namespace with application? I don't think we are. Which means that:

1) these comments are wrong
2) pargs need to include the application as a criterion

I think I need to think this through thoroughly and come up with a better appID story.

ZachEddy: > I think I need to think this through thoroughly and come up with a better appID story.

Does that mean we should we hold off on this pull request altogether?

johnbelamaric: ok. that works for now. I will write something up on Monday, I have a few thoughts around how we can make this all work better.

markinos:
Lgtm, although since you are screening for upper case letters in namespaces, note that kubernetes also restricts punctuation to just dashes - and periods .

ZachEddy:
Thanks, Mark! The `appID` just corresponds to a Kubernetes namespace. Nothing will break if the `appID` isn't a valid namespace, but the authorization interceptor will deny every request. Shouldn't be an issue.

PR #35, opened 2018-05-18
Server struct

dmacthedestroyer:
Implement a `Server` struct that wraps the GRPC server, gateway and health checks.
Has a variadic `Option` constructor, so more options can be added. Some examples:
* logger
* metrics

After creating the `Server`, you can start it with the `.Serve(net.Listener, net.Listener)` function that takes a listener for the grpc server and http server, respectively.

**Why two listeners?** I tried getting it all in one port, but ran into many issues across the tcp layer through the application layer. Every workaround I found had caveats and, in the end, I decided that two ports was the least-ugly implementation I could muster.

**To support TLS**, wrap your `net.Listeners` with `tls.NewListener(net.Listener, tls.Config)` before calling `.Serve`

This PR also has stuff for flags in it, which isn't directly related to the rest of the work, and needs further work. I've been dragging this PR on for too long, though, so I wanted to get it out for review.


server/server.go
```diff
@@ -0,0 +1,214 @@
...
+			if err := s.HTTPServer.Close(); err != nil {
+				errC <- err
+			}
+		}
+	}()
+
+	select {
+	case err := <-errC:
+		return err
+	case <-doneC:
```
ZachEddy: Disclaimer: I'm not a concurrency expert, but I don't think is quite right. Shouldn't the `select` continue until both the HTTP server _and_ the gRPC are closed? In this case, the first server that closes will trigger the `return` statement.

If this isn't the behavior you want, then I think [buffered channels](https://tour.golang.org/concurrency/3) will come in handy.

servertest/servertest.go
```diff
@@ -0,0 +1,16 @@
+package servertest
+
+import (
+	"net"
+)
+
+// NewLocalListener returns a system-chosen port on the local loopback interface, for use in end-to-end HTTP tests
+func NewLocalListener() (net.Listener, error) {
```
ZachEddy: Why not include `NewLocalListener` inside the `server` package? Is this intended to be a utility that developers can use for their own testing purposes?

If so, adding some documentation to the toolkit README that describes the `server` and `servertest` packages would be helpful (once the PR is closer to getting merged).

server/server.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"errors"
+
+	"github.com/infobloxopen/atlas-app-toolkit/gateway"
+	"github.com/infobloxopen/atlas-app-toolkit/health"
+	"google.golang.org/grpc"
+)
+
+var (
+	ErrInitializeTimeout      = errors.New("initialization timed out")
```
ZachEddy: This got left out of my earlier review, but these should have comments.

Running `golint` turned this up and a few others:

```
flag.go:11:6: exported type GRPCFlags should have comment or be unexported
flag.go:15:6: exported type TLSFlags should have comment or be unexported
flag.go:21:6: exported type HealthProbesFlags should have comment or be unexported
flag.go:27:1: exported function NewHealthProbesFlags should have comment or be unexported
flag.go:37:1: exported method HealthProbesFlags.Addr should have comment or be unexported
flag.go:41:1: exported method HealthProbesFlags.HealthPath should have comment or be unexported
flag.go:45:1: exported method HealthProbesFlags.ReadyPath should have comment or be unexported
flag.go:49:1: exported function MetricsFlags should have comment or be unexported
flag.go:55:1: exported function NewGRPCFlags should have comment or be unexported
flag.go:61:1: exported method GRPCFlags.Addr should have comment or be unexported
flag.go:65:1: exported function NewTLSFlags should have comment or be unexported
flag.go:74:1: exported method TLSFlags.TLSConfig should have comment or be unexported
flag.go:78:1: exported method TLSFlags.WithGRPCTLSCreds should have comment or be unexported
server.go:21:2: exported var ErrInitializeTimeout should have comment or be unexported
server.go:39:6: exported type Option should have comment or be unexported
server.go:41:6: exported type InitializerFunc should have comment or be unexported
server.go:161:1: exported method Server.Stop should have comment or be unexported
tls.go:1:1: package comment should be of the form "Package server ..."
```

server/server.go
dmacthedestroyer: This is what I wanted -- if either of the servers go down, then the entire thing should go down. That way you're not running a gateway without the gRPC server, or vice-versa.

servertest/servertest.go
dmacthedestroyer: good callout on updating the README.  I copied what Go did with their [httptest package](https://golang.org/pkg/net/http/httptest/) to get this. It's a test helper thing, so it shouldn't live in the non-test `server` package.

server/server.go
dmacthedestroyer: good call -- I need to figure out how to get linter errors in my IDE!

server/tls.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	roots, err := loadRoots(caPath)
+	if err != nil {
+		return nil, err
+	}
+
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		RootCAs:      roots,
```
johnbelamaric: For server certs, this should be ClientCAs, not RootCAs. Yes, this is a bug in the CoreDNS TLS code, I believe. I will probably need to fix that to get client-based cert working there.

johnbelamaric: Also make sure this work in combination with a hand-created tls.Listener - I think it should if you use the same tlsConfig for both. But I seem to recall when I was playing with it that I couldn't get client-certs to work with using my own tls Listener.

server/server.go
```diff
@@ -0,0 +1,214 @@
...
+			_, err := gateway.NewGateway(append(options, gateway.WithMux(mux))...)
+			return err
+		})
+		return nil
+	}
+}
+
+// Serve invokes all initializers then serves on the given listeners.
+//
+// If a listener is left blank, then that particular part will not be served.
```
amaskalenka: If both part are `nil` then we will serve nothing until forever?

server/server.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	return <-errC
+}
+
+func (s *Server) Stop() error {
+	doneC := make(chan bool)
+	errC := make(chan error)
+	go func() {
+		defer func() { doneC <- true }()
```
amaskalenka: Why do you actually need a deferred call here?

server/server.go
```diff
@@ -0,0 +1,214 @@
...
+		if s.GRPCServer != nil {
+			s.GRPCServer.Stop()
+		}
+	}()
+	go func() {
+		defer func() { doneC <- true }()
```
amaskalenka: and here

server/server.go
```diff
@@ -0,0 +1,214 @@
...
+		errC <- nil
+	}()
+
+	for _, i := range s.initializers {
+		go func() {
+			defer wg.Done()
+			if err := i(ctx); err != nil {
+				errC <- err
+			}
+		}()
```
amaskalenka: It seems to me that `i` should be passed as an argument.

gateway/gateway.go
```diff
@@ -59,9 +60,9 @@ func (g gateway) registerEndpoints() (*http.ServeMux, error) {
 			}
 		}
 		// strip prefix from testRequest URI, but leave the trailing "/"
-		mux.Handle(prefix, http.StripPrefix(prefix[:len(prefix)-1], gwmux))
+		g.mux.Handle(prefix, http.StripPrefix(prefix[:len(prefix)-1], gwmux))
```
amaskalenka: I am not sure I really understand why we need to leave trailing "/" here and how is it validated that it will be exactly "/"?
What if an user pass `/v1` as prefix? Will the `gmux` receive `/v`?
Could you please add comment that explain such need in order to help someone who probably will work on this code in some day?

server/server.go
dmacthedestroyer: good find! 
the `http.Server` struct also has both [`Shutdown`](https://golang.org/pkg/net/http/#Server.Shutdown) and [`Close`](https://golang.org/pkg/net/http/#Server.Close) functions, so I could build out the two functions for both services

server/server.go
dmacthedestroyer: you're right: I probably don't. I just blindly followed the concurrency patterns I found with wait groups. 

I kind of like it like this, though: in general it means that you can do whatever the hell you want below the line and still be sure that your `wg.Done()` gets called. I'm gonna keep it like this :)

server/server.go
dmacthedestroyer: good call

gateway/gateway.go
dmacthedestroyer: this is a good callout, and should probably be changed. There also weren't any tests associated with this `gateway.go` file, so more work to do there

PR #36, opened 2018-05-18
Add WithTLS

johnbelamaric:
This will be used along with mutual TLS to provide authorization for service-to-service requests.

### DIFF
--- a/auth/options.go
+++ b/auth/options.go
@@ -9,6 +9,8 @@ import (
 	jwt "github.com/dgrijalva/jwt-go"
 	pdp "github.com/infobloxopen/themis/pdp-service"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
 )
 
 // functional options for the defaultBuilder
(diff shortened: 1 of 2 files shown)


auth/options.go
```diff
@@ -93,9 +95,44 @@ func getRequestDetails(ctx context.Context) (string, string, error) {
...
+	f := func(ctx context.Context) ([]*pdp.Attribute, error) {
+		p, ok := peer.FromContext(ctx)
+		if !ok || p.AuthInfo == nil {
+			return []*pdp.Attribute{
+				{Id: tlsVerified, Type: "boolean", Value: "false"},
+			}, nil
+		}
+
+		t, ok := p.AuthInfo.(*credentials.TLSInfo)
+		if !ok || t.State.VerifiedChains == nil || len(t.State.VerifiedChains) == 0 || len(t.State.VerifiedChains[0]) == 0 {
```
ZachEddy: Can `t.State` ever be `nil` here? If so, there should be a `nil` check on `t.State` before referencing the `VerifiedChains` field.

johnbelamaric: t.State is a struct

PR #38, opened 2018-05-18
Transaction management

amaskalenka:
- Added: Transaction wrapper around `gorm.DB`
- Added: UnaryServerInterceptor to manage Transaction wrapper

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,9 @@ The created transaction is stored in `context.Context` and passed to the request
       ),
     ),
   )
+```
 
+```go
 import (
 	"github.com/infobloxcopen/atlas-app-toolkit/gorm"
 )
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ For this purpose `auth.GetAccountID(ctx, nil)` function can be used:
 func (s *contactsServer) Read(ctx context.Context, req *ReadRequest) (*ReadResponse, error) {
 	input := req.GetContact()
 
-	accountID, err := mw.GetAccountID(ctx, nil)
+	accountID, err := auth.GetAccountID(ctx, nil)
 	if err == nil {
 		input.AccountId = accountID
 	} else if input.GetAccountId() == "" {
@@ -322,7 +322,7 @@ func (s *myServiceImpl) MyMethod(ctx context.Context, req *MyRequest) (*MyRespon
 }
 ```
 
-Also you can use our helper function `gw.Header()`
+Also you can use our helper function `gateway.Header()`
 
 ```golang
 import (
@@ -334,7 +334,7 @@ import (
 func (s *myServiceImpl) MyMethod(ctx context.Context, req *MyRequest) (*MyResponse, error) {
     var userAgent string
 
-    if h, ok := gw.Header(ctx, "user-agent"); ok {
+    if h, ok := gateway.Header(ctx, "user-agent"); ok {
         userAgent = h
     }
 }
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -83,6 +83,41 @@ func (s *contactsServer) Read(ctx context.Context, req *ReadRequest) (*ReadRespo
...
+    ),
+  )
+
+import (
+	"github.com/infobloxcopen/atlas-app-toolkit/gorm"
+)
+
+func (s *MyService) MyMethod(ctx context.Context, req *MyMethodRequest) (*MyMethodResponse, error) {
+	// extract gorm transaction from context
+	txn, ok := gorm.FromContext(ctx)
```
Nickolai-Belov: (minor) Guess it would be nice to have more self-explaining names for these funcs.
E.g. ` transaction.NewContext( )` or `gorm.NewTransactionContext( )` instead of `gorm.NewContext( )`

gorm/transaction.go
```diff
@@ -0,0 +1,124 @@
...
+		defer func() {
+			if terr == nil {
+				return
+			}
+			st := status.Convert(err)
+			st, _ = st.WithDetails(errdetails.New(codes.Internal, "gorm", terr.Error()))
+			err = st.Err()
+		}()
+
+		if terr = txn.Begin().Error; terr != nil {
```
Nickolai-Belov: Does it mean that transaction is always opened even if not needed in the request?

README.md
amaskalenka: That is common naming scheme for context related functions.
Strictly speaking transaction is a `gorm.DB` instance and `gorm.NewContext()` with opposite `gorm.FromContext()`  looks convenient and self descriptive FMPOV.

I intentionally did not add separate `transaction` package because it requires definition of generic `Transaction` interface that performs transaction management and does not depend on any of underlying implementation (sql.Tx, gorm, etc..)

gorm/transaction.go
amaskalenka: Yes.

gorm/transaction.go
```diff
@@ -0,0 +1,124 @@
...
+// It works as a singleton to prevent an application of creating more than one
+// transaction instance per incoming request.
+type Transaction struct {
+	mu      sync.Mutex
+	parent  *gorm.DB
+	current *gorm.DB
+}
+
+// Begin starts new transaction by calling `*gorm.DB.Begin()`
+// Returns new instance of `*gorm.DB` (error can be checked by `*gotm.DB.Error`)
```
Evgeniy-L: Spelling error - gorm.DB.Error

README.md
```diff
@@ -83,6 +83,41 @@ func (s *contactsServer) Read(ctx context.Context, req *ReadRequest) (*ReadRespo
...
+    grpc.UnaryInterceptor(
+      grpc_middleware.ChainUnaryServer( // middleware chain
+        ...
+        gorm.UnaryServerInterceptor(), // transaction management
+        ...
+      ),
+    ),
+  )
+
+import (
```
Evgeniy-L: does it make sense to have code in two separate blocks instead of one?

gorm/transaction.go
amaskalenka: Thanks.

gorm/transaction.go
Nickolai-Belov: Why is it needed here? The transaction may be opened on demand in the service implementation as shown in the [readme](https://github.com/infobloxopen/atlas-app-toolkit/pull/38/files#diff-04c6e90faac2675aa89e2176d2eec7d8R116).

PR #45, opened 2018-05-29
Add Request-Id middleware

kumaya:
- Add interceptor to extract Request-Id from incoming metadata and update
    context and outgoing metadata with the same.
  - Create new Request-Id in cases when incoming metadata does not contain
    Request-Id
  - Add unit tests


requestid/interceptor.go
```diff
@@ -0,0 +1,37 @@
...
+		errd := UpdateHeader(newCtx, reqId)
+		if errd != nil {
+			errd = status.Errorf(codes.Internal, "request id interceptor: %s", errd)
+			grpclog.Errorln(errd)
+		}
+		return
+	}
+}
+
+func FromContext(ctx context.Context) (exists bool, reqId string) {
```
Evgeniy-L: I would prefer to have method look like `FromContext(ctx context.Context) (reqId string, exists bool)`
E.g. return params in different order. This way it looks consistent with other go code (see `reqId, exists = ctx.Value(requestIDKey{}).(string)` below)

Evgeniy-L:
Overall looks good to me

requestid/interceptor.go
amaskalenka: I think you need add function `NewContext` as an opposite one to the `FromContext` (it may be useful in case if someone wants to override RequestID for whatever reason)

requestid/requestid.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+	"github.com/google/uuid"
+	"github.com/infobloxopen/atlas-app-toolkit/gateway"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// DefaultRequestIDKey is the metadata key name for request ID
+var DefaultRequestIDKey = "Request-Id"
```
amaskalenka: How can I override `default` key? I see only one  way is to assign new value to the global public variable. Is it correct?

requestid/interceptor_test.go
```diff
@@ -0,0 +1,69 @@
+package requestid
+
+import (
+	"context"
```
amaskalenka: Please use `goimports`

requestid/requestid.go
```diff
@@ -0,0 +1,36 @@
+package requestid
+
+import (
+	"context"
+	"github.com/google/uuid"
```
amaskalenka: I do not see `Godep` files updated.

requestid/requestid.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/infobloxopen/atlas-app-toolkit/gateway"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// DefaultRequestIDKey is the metadata key name for request ID
+var DefaultRequestIDKey = "Request-Id"
+
+func HandleRequestId(ctx context.Context) (reqId string) {
```
amaskalenka: Please add documentation to the public functions you added. (use `golint` to make sure your code in desirable quality)

requestid/interceptor.go
```diff
@@ -0,0 +1,37 @@
...
+		reqId := HandleRequestId(ctx)
+
+		newCtx := context.WithValue(ctx, requestIDKey{}, reqId)
+
+		// returning from the request call
+		res, err = handler(newCtx, req)
+
+		errd := UpdateHeader(newCtx, reqId)
+		if errd != nil {
+			errd = status.Errorf(codes.Internal, "request id interceptor: %s", errd)
```
amaskalenka: Please provide more descriptive error message for an internal error in order to make debugging less painful.

requestid/interceptor.go
```diff
@@ -0,0 +1,37 @@
...
+type requestIDKey struct{}
+
+func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (res interface{}, err error) {
+		reqId := HandleRequestId(ctx)
+
+		newCtx := context.WithValue(ctx, requestIDKey{}, reqId)
+
+		// returning from the request call
+		res, err = handler(newCtx, req)
```
amaskalenka: What if `handler` throws `panic`?

kumaya:
@amaskalenka: Each request need to be assigned with its own request_id, so when the request goes from one service to another it can be easily traced in logs. This interceptor plans to extract Request-Id from incoming metadata and update context with the same.
Can you suggest a name that suits well ?

requestid/requestid.go
kumaya: https://github.com/infobloxopen/atlas-app-toolkit/blob/master/Gopkg.lock#L38

amaskalenka:
In the code I see that in case if an incoming request does not have `request_id` the new one is generated and sends back to the caller. What is the reason for it (send it back)?
```go
// SetHeader sets the header metadata.
// When called multiple times, all the provided metadata will be merged.
// All the metadata will be sent out when one of the following happens:
//  - grpc.SendHeader() is called;
//  - The first response is sent out;
//  - An RPC status is sent out (error or success).
```

According to the your description the original intention is to track/trace requests, so I think it make sense to reflect that in the package name.

kumaya:
@amaskalenka: For cases where a gRPC service is invoking other services and does not contain request-id. For those scenarios as well we need tracing the flow. So, we are sending it back. So that services like audit-logging can map the request flow.

amaskalenka:
@kumaya I do not see how `outgoing gRPC metadata` is related to separate service call.
`service A -sends request A (as part of handling request (0) to itself) -> service B -and service B responds with request id attached`
How does `service A` couple request id from (0) to request B?

Can you provide more detailed use case?

dmacthedestroyer:
This seems like a stopgap for Zipkin, likely because the Audit Logging project needs this now. Would it be worth it investigate Zipkin integration instead of building this middleware? At the very least, this should be able to be interchangeable with whatever we come up with for tracing.

kumaya:
@amaskalenka:
The logic of creation of Request-Id is of following form:
	a. If metadata contains Request-Id, interceptor check metadata from incoming context; finds one and simply update the context with Request-Id.
	b. If metadata does not contain Request-Id, interceptor creates one and updates the context and metadata with Request-Id.

The flow of request can be of the following form across service 0, service A and service B:
	service 0 does not have request-id header: In such cases, requestid interceptor in service 0 check for header in incoming metadata, since Request-Id is absent, it creates one and updates the context and metadata. When service 0 calls service A (either thru gRPC or gateway with its context), requestid interceptor (of service A) checks for Request-Id in metadata, finds one (since requestid interceptor in service 0 updated the context and metadata already) and passes the flow to handler. Same happens with flow from service A to service B.

Please add if this can be improved. One way is to not bringing the headers at all and only doing `FromContext` and `NewContext` to include Request-Id.

requestid/requestid.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	return reqID
+}
+
+func newRequestID() string {
+	return uuid.New().String()
+}
+
+func updateHeader(ctx context.Context, reqID string) error {
```
amaskalenka: You should delete this function.

requestid/requestid.go
```diff
@@ -0,0 +1,37 @@
...
+	md := metadata.Pairs(DefaultRequestIDKey, reqID)
+	return grpc.SetHeader(ctx, md)
```
amaskalenka: You set headers for response here!

requestid/interceptor.go
```diff
@@ -0,0 +1,61 @@
...
+	if !exists {
+		return "", false
+	}
+	return reqID, true
+}
+
+// NewContext creates a new context with Request-Id attached.
+func NewContext(ctx context.Context, reqID string) context.Context {
+	return context.WithValue(ctx, requestIDKey{}, reqID)
+}
```
amaskalenka: add 
```go
md := metadata.Pairs(DefaultRequestIDKey, reqID)
return metadata. AppendToOutgoingContext(ctx, md)
```

kumaya: `AppendToOutgoingContext` does not take `md` instead take kv pair. If the idea is to override the existing `request-id` we should use `NewOutgoingContext(ctx context.Context, md MD)`
https://github.com/grpc/grpc-go/blob/master/metadata/metadata.go#L155

amaskalenka: You are right.

PR #48, opened 2018-06-01
NGP-916

askurydzin:
The root cause of the bug was runtime.Marshaller limited capabilities and it's inability to marshal nil values inside map, resulting in following error:

```
panic: reflect: call of reflect.Value.Interface on zero Value [recovered]
	panic: reflect: call of reflect.Value.Interface on zero Value
```

This fix implements logic for response map traversal and nil value removal. 

```
$ make test
test -z `docker run --privileged --rm -v /home/askurydzin/go/src/github.com/infobloxopen/atlas-app-toolkit:/go/src/github.com/infobloxopen/atlas-app-toolkit -w /go/src/github.com/infobloxopen/atlas-app-toolkit infoblox/buildtool:v8 go fmt ./...`
docker run --privileged --rm -v /home/askurydzin/go/src/github.com/infobloxopen/atlas-app-toolkit:/go/src/github.com/infobloxopen/atlas-app-toolkit -w /go/src/github.com/infobloxopen/atlas-app-toolkit infoblox/buildtool:v8 go test ./...
ok  	github.com/infobloxopen/atlas-app-toolkit/auth	0.020s
ok  	github.com/infobloxopen/atlas-app-toolkit/gateway	0.020s
ok  	github.com/infobloxopen/atlas-app-toolkit/gorm	0.005s
ok  	github.com/infobloxopen/atlas-app-toolkit/health	1.049s
ok  	github.com/infobloxopen/atlas-app-toolkit/query	0.009s
ok  	github.com/infobloxopen/atlas-app-toolkit/requestid	0.005s
ok  	github.com/infobloxopen/atlas-app-toolkit/rpc/errdetails	0.001s
ok  	github.com/infobloxopen/atlas-app-toolkit/server	0.119s
?   	github.com/infobloxopen/atlas-app-toolkit/servertest	[no test files]
```

### DIFF
--- a/gateway/response.go
+++ b/gateway/response.go
@@ -91,6 +91,12 @@ func (fw *ResponseForwarder) ForwardMessage(ctx context.Context, mux *runtime.Se
 
 	retainFields(ctx, req, dynmap)
 
+	// FIXME: standard grpc JSON marshaller doesn't handle
+	// nil values inside maps.
+	for k := range dynmap {
+		dynmap[k] = fixNilValues(dynmap[k])
+	}
+
 	// Here we set "Location" header which contains a url to a long running task
 	// Using it we can retrieve its status
 	rst := Status(ctx, nil)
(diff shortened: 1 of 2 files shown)


dmacthedestroyer:
is there a formal fix pending in the grpc-gateway repo? I'd recommend at the very least opening an issue for this if the problem is with their marshaller, and possibly making a PR for the fix as well.

askurydzin:
i guess it will be hard to propagate the fix to grpc-gateway because it would inflame a holy war regarding returning nillable values as null or not return them at all. In our case, northstar would like to use wkt to distinguish between 0 and empty value, however their main purpose is nillable option. marshaller in grpc-gateway was not intended to do something more complicated than map of ints, strings and enums, and the main proposal would be to write our own marshaller.

dmacthedestroyer:
This [PR](https://github.com/grpc-ecosystem/grpc-gateway/pull/242) seems to address this same issue, if I'm not mistaken.  Looks like there's support for it but no one is willing to fix tests :)

PR #50, opened 2018-06-01
Implement intercepter for validating parameter dependencies

ystankevich:
In NGP-913, Arnaud Klein requested "An error should be thrown when _offset and/or _limit are used without _order_by".
In "REST API Syntax Specification" I didn't find such dependencies, and it looks like non common.
So do we need to add such interceptor into toolkit? or service developers should implement it for their needs?

### DIFF
--- a/gateway/operator.go
+++ b/gateway/operator.go
@@ -18,14 +18,14 @@ import (
 const (
 	filterQueryKey           = "_filter"
 	filterMetaKey            = "operator-filter"
-	sortQueryKey             = "_order_by"
-	sortMetaKey              = "operator-sort"
+	SortQueryKey             = "_order_by"
+	SortMetaKey              = "operator-sort"
 	fieldsQueryKey           = "_fields"
 	fieldsMetaKey            = "operator-fields"
-	limitQueryKey            = "_limit"
-	limitMetaKey             = "operator-limit"
-	offsetQueryKey           = "_offset"
-	offsetMetaKey            = "operator-offset"
+	LimitQueryKey            = "_limit"
+	LimitMetaKey             = "operator-limit"
+	OffsetQueryKey           = "_offset"
+	OffsetMetaKey            = "operator-offset"
 	pageTokenQueryKey        = "_page_token"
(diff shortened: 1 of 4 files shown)


johnbelamaric:
The easier way is to have _order_by internally always append the primary key to the sort order list. Really not only is it the easiest way, but it's the only way to guarantee a stable sort order for paging.

johnbelamaric:
Sure, they can add this if they need it. I guess having it as a separate middleware is fine.

My point is that our generated gorm code should always sort by PK. If they ask for a specific sort, then we put those keys in front of PK. If you don't do that, you are not guaranteed a stable sort order. Imagine if they sort on "first name" in a database with 1M names. In order to page properly, we need that sort to be stable, but that is not guaranteed by the database. So, they ask for sort of "first name" and we sort by "first name, PK". Then you have a stable sort order.

gateway/operator.go
```diff
@@ -18,14 +18,14 @@ import (
...
-	sortMetaKey              = "operator-sort"
+	SortQueryKey             = "_order_by"
+	SortMetaKey              = "operator-sort"
 	fieldsQueryKey           = "_fields"
 	fieldsMetaKey            = "operator-fields"
-	limitQueryKey            = "_limit"
-	limitMetaKey             = "operator-limit"
-	offsetQueryKey           = "_offset"
-	offsetMetaKey            = "operator-offset"
+	LimitQueryKey            = "_limit"
```
amaskalenka: I am not sure that a good idea make them public.

johnbelamaric:
...and if we do that with our gorm code, then this code is unnecessary, because we always return data in PK order if they don't specify an order.

Of course, if they are not using our gorm code then they either have to do the equivalent of what we are doing, or have this middleware. But as stated above, just having an order by does not guarantee a stable sort for paging, so I think it's providing a false sense of security.

dmacthedestroyer:
I guess I was being too abstract in my thinking, and I'm on board with what you're saying. I'd also prefer to not even include this middleware, so if we add work to ensure an order by clause for `List` functions then maybe we can close this PR?

PR #53, opened 2018-06-13
Implement interceptor which parse collection operators and stores res…

ystankevich:
…ult in context

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -2,8 +2,6 @@ package gateway
 
 import (
 	"context"
-	"fmt"
-	"reflect"
 
 	"github.com/infobloxopen/atlas-app-toolkit/query"
 	"google.golang.org/grpc"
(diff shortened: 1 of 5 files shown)


ystankevich:
That allow in future, to modify collection operators  for developers, store them in context and underlying code will use that new collection operators.

gateway/middleware.go
```diff
@@ -29,143 +34,104 @@ func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+//
+// Collection operators setters
+//
+func CtxSetFiltering(ctx context.Context, f *query.Filtering) context.Context {
+	return context.WithValue(ctx, filteringCollection, f)
+}
 
-	for i := 0; i < reqval.NumField(); i++ {
-		f := reqval.FieldByIndex([]int{i})
+func CtxSetSorting(ctx context.Context, s *query.Sorting) context.Context {
```
Evgeniy-L: Please add proper comments to methods.

dmacthedestroyer:
What's the intention for this PR? It looks like you're trying to store the collection operators in context to be able to modify it through the middleware chain. 

Does this make it so that the collection operators in the protobuf request message can become out of sync with what's in the context?

ystankevich:
That is fix for NGP-1019.
In that fix implemented functions which allow to get collection parameters from context.
Also allow to create new context with new collection parameters.
So developer may modify or use specific collection parameters even if they are not provided by initial request

ystankevich:
We will partly use that for supporting filtering by synthetic fields.

dmacthedestroyer:
My main concern is that if someone puts a collection operator in their request message proto, there's a possibility that because of this middleware the value provided to their service implementation will be different from what's being stored in the context. 

Is there any assurance that these two places won't get out of sync?

PR #54, opened 2018-06-14
Resource implementation

amaskalenka:
Protocol Buffer representation.
JSONPB implementation
Resource management: registration, building, resolving
Default implementation of UUID resource management.
Default implementation of External resource management.


rpc/resource/resource.go
```diff
@@ -0,0 +1,102 @@
...
+type Identifier interface {
+	fmt.Stringer
+	sql.Scanner
+	driver.Valuer
+
+	ApplicationName() string
+	ResourceType() string
+	ResourceID() string
+
+	IsExternal() bool
```
amaskalenka: I do really want a feedback on `Identifier` interface. Is not it overloaded?

amaskalenka: It seems that we could narrow interface by eliminating ‘sql.Scanner’, ‘fmt.Stringer’ and ‘driver.Valuer’ interfaces, thereby allow an user to decide is an Identifier is storable or not.

Evgeniy-L: I am fine with both ways. Probably the idea to eliminate  `sql.Scanner`, `fmt.Stringer` and `driver.Valuer` interfaces makes sence as it might be some service without DB at all.

rpc/resource/internal/string.go
```diff
@@ -0,0 +1,45 @@
...
+func ParseString(id string) (aname, rtype, rid string) {
+	v := strings.SplitN(strings.Trim(id, Delimiter), Delimiter, 3)
+	switch len(v) {
+	case 1:
+		rid = v[0]
+	case 2:
+		rtype, rid = v[0], v[1]
+	case 3:
+		aname, rtype, rid = v[0], v[1], v[2]
+	}
```
akleinib: Hopefully no one will ever put a / in a `resource_id`.

rpc/resource/resource.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	v, ok := registry[name]
+	if !ok || v == nil || v.Manager == nil {
+		return nil, fmt.Errorf("resource: manager is not registered for resource %s", name)
+	}
+
+	return v.Manager.Build(identifier)
+}
+
+func Resolve(message proto.Message, identifier Identifier) (*resourcepb.Identifier, error) {
```
akleinib: Isn't it an entire cut'n'paste of `Build()` except for the last line?

amaskalenka: Yes, it is.

rpc/resource/resource.go
```diff
@@ -0,0 +1,114 @@
...
+	// Encode returns RPC representation of Identifier interface
+	Encode(Identifier) (*resourcepb.Identifier, error)
+	// Decode returns implementation of Identifier interface based on RPC representation
+	Decode(*resourcepb.Identifier) (Identifier, error)
+}
+
+// RegisterCodec registers codec for a given pb.
+// If pb is nil the codec is registered as default.
+// If codec is nil or registered twice for the same resource
+// the panic is raised.
```
dmacthedestroyer: why panic instead of return `error`?

rpc/resource/resourceext/resource.go
```diff
@@ -0,0 +1,103 @@
+package resourceext
```
dmacthedestroyer: could this just live in the `resource` package? What are the benefits of putting it inside another package?

amaskalenka: This is common pattern to separate interface from implementations. See [package image](https://godoc.org/image) or [hash package](https://godoc.org/hash)  for example.

rpc/resource/resource.go
amaskalenka: Registration happens at start of an application and should be defined somewhere in the code e.g in `init` functions. If an developer is registering the same codec twice for the same resource it seems it a bug.

It is done to indicate that a codec is overrided. 

It is unlikely that someone would want to encode/decode identifiers at runtime using one codec and then after some period change the codec to another one.

dmacthedestroyer: one idea if it should happen at application startup only is that it is only accessible via a constructor instead of as a function. That way it's impossible to abuse the functionality.

Also, I see you're using a global singleton pattern here, which is somewhat concerning... I guess Go does it liberally as well, but I'm struggling to not see it as an antipattern.

rpc/resource/external/resource.go
```diff
@@ -0,0 +1,113 @@
+package external
```
dmacthedestroyer: please add comments to explain the package. `external` is easy to misunderstand without full context

rpc/resource/internal/string.go
```diff
@@ -0,0 +1,45 @@
...
+	return strings.Join(l, Delimiter)
+}
+
+// ParseString parses id according to Atlas Reference format:
+//	<application_name>/<resource_type>/<resource_id>
+// All leading and trailing Delimiter are removed.
+// The resource_id is parsed first, then resource type and last application name.
+// The id "/a/b/c/" will be converted to "a/b/c" and returned as (a, b, c).
+// The id "b/c/" will be converted to "b/c" and returned as ("", b, c).
+func ParseString(id string) (aname, rtype, rid string) {
```
dmacthedestroyer: it might be useful to publicly export this, in cases where people are storing references as a string instead of our custom type

rpc/resource/external/resource.go
```diff
@@ -0,0 +1,113 @@
...
+	return i.String(), nil
+}
+
+func (i *identifier) Scan(v interface{}) error {
+	if v == nil {
+		return nil
+	}
+	s, ok := v.(string)
+	if !ok {
+		return fmt.Errorf("external: invalid sql type of resource id %T", v)
```
dmacthedestroyer: hmm... this interface has hidden implications. Should it be specific to `sql` engines? I feel like if it's for sql drivers, it should live in a `sql` package, not an `external` package.

rpc/resource/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+			return nil, err
+		} else {
+			v.ExternalID = id
+		}
+		return &v, nil
+	}
+
+	// let's create PB message
+	pb := &ExampleProtoMessage{
+		Id: &resourcepb.Identifier{
```
dmacthedestroyer: could this be left `nil`?

rpc/resource/external/resource.go
amaskalenka: That is implementation of `sql.Scanner` interface, and yes it is sql specific.

rpc/resource/internal/string.go
amaskalenka: it might be, but let's keep it private for now. It is easy to make functions public than private.

rpc/resource/resource.go
amaskalenka: Ok, what do you propose?

rpc/resource/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+			ApplicationName: "",
+			// resource type is empty as well
+			ResourceType: "",
+			// resource id is empty because we want to create new TestProtoMessage
+			ResourceId: "",
+		},
+		// ExternalId stores data about "external_resource" that belongs to
+		// "externalapp" and has id "1"
+		ExternalId: &resourcepb.Identifier{
+			ApplicationName: "externalapp",
```
dmacthedestroyer: so is the idea that these reference representations should only be used when you're storing any kind of object from any king of application/resource? 

If this example app has a reference to a field that is for a specific application, then the `ApplicationName` shouldn't be allowed to be entered by users. In that case, they can just use a `string` or `int` field to represent it. Correct?

rpc/resource/example_test.go
amaskalenka: yes

amaskalenka:
My `golint` and `gometalinter` are happy.

rpc/resource/resource.go
dmacthedestroyer: don't know yet... just throwing out unqualified opinions at this point :)

Evgeniy-L:
Overall looks good to me. One question is why do we add so many packages?
 - rpc/resource
 - rpc/resource/fq
 - rpc/resource/internal
 - rpc/resource/resourcepb
 - rpc/resource/uuid
Most of them contain only two files (x.go and x_test.go)

amaskalenka:
The `resource` package contains top interfaces and functions that work with those interfaces.
The `resource/uuid` and `resource/fq` contain implementations of top level interfaces. (like `image` package from std library contains `png` and `jpeg` implementations)
The `resource/resourcepb` contains generated PB definition and named in that way to separate it from implementations.

rpc/resource/uuid/resource_test.go
```diff
@@ -0,0 +1,240 @@
...
+		{
+			Value:      "uuid",
+			Valid:      true,
+			ResourceID: "uuid",
+		},
+		{
+			Value: nil,
+		},
+		{
+			Value:         2018,
```
Nickolai-Belov: So to add resource identifier support for incremental PK one should implement Identifier and Codec separately, right?

rpc/resource/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+# How to use in Golang?
+
+The common interfaces and helper functions to work with identifiers are defined in [resource](resource.go) file.
+There are also some common implementations. 
+
+The [fq](fq/resource.go) package provides a codec to encode and decode [Protocol Buffer representation](resourcepb/resource.proto) of 
+identifiers to and from `Identifier`.
+The implementation of `identifier` supports `sql.Scanner` and `driver.Valuer` interfaces
+so it could be stored in SQL DB as a single string delimited by `/` (in **f**ully **q**ualified from).
```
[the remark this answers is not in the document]
amaskalenka: 👍

rpc/resource/uuid/resource_test.go
amaskalenka: Right!

rpc/resource/resource.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package infoblox.rpc;
```
Calebjh: Should this be infoblox.rpc or atlas.rpc?

amaskalenka: Hmm..., I don't know

johnbelamaric:
atlas is better

rpc/errdetails/error_details.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package infoblox.rpc;
+package atlas.rpc;
```
Evgeniy-L: this may effect recent changes for error handling. this need to be properly tested!

rpc/resource/README.md
```diff
@@ -0,0 +1,33 @@
...
+```proto
+syntax = "proto3";
+
+import "github.com/infobloxopen/atlas-app-toolkit/rpc/resource/resource.proto";
+
+message MyMessage {
+    infoblox.rpc.Identifier id = 1;
+    string some_data = 2;
+    infoblox.rpc.Identifier reference_on_external_resource = 3;
+    infoblox.rpc.Identifier foreign_id_of_internal_resource = 4;
```
Evgeniy-L: atlas.rpc.Identifier

gorm/resource/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+option go_package = "github.com/yourapp/pb;pb";
+
+message A {
+    option (gorm.opts).ormable = true;
+    
+    infoblox.rpc.Identifier id = 1 [(gorm.field).tag = {type: "integer"}];
+    string value = 2;
+    repeated B b_list = 3; // has many
+    infoblox.rpc.Identifier external = 4 [(gorm.field).tag = {type: "text"}];
```
Evgeniy-L: atlas.rpc.Identifier

gorm/resource/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+The common interfaces, types and helper functions to work with identifiers are defined in [resource](resource.go) file.
+
+There are also some common implementations. 
+
+# Codecs
+
+The [fqstring](fqstring/codec.go) package provides a codec to encode and decode [atlas.rpc.Identifier](../../rpc/resource/resource.proto) 
+to and from `driver.Value`.
+The implementation converts `infoblox.rpc.Identifier` to a single string delimited by `/` (in **f**ully **q**ualified form).
```
Evgeniy-L: atlas.rpc.Identifier (multiple places)

Evgeniy-L:
Overall LGTM but please make sure errors functionality is not effected by `infoblox.rpc` -> `atlas.rpc` change. Also is we do this change we should mention that we added a backward incompatible change in release notes.

PR #55, opened 2018-06-15
Add BeginFromContext for transaction handling, and tests

Calebjh:
Add a helper function for pulling transaction handler from context and beginning transaction.

### DIFF
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -21,11 +21,10 @@ type ctxKey int
 // instead of using this key directly.
 var txnKey ctxKey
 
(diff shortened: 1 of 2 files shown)


gorm/transaction.go
```diff
@@ -20,6 +21,12 @@ type ctxKey int
 // instead of using this key directly.
 var txnKey ctxKey
 
+type CtxTxnMissingError struct{}
```
amaskalenka: Did you add new type to provide a way for developer to distinguish transaction error?
If so, I would change it on public variable instead.

PR #57, opened 2018-06-20
Created a new middleware that parses through the validation errors

razamiDev:
Created a new middleware that parses through the validation errors and returns a user friendly error. Created unit tests to ensure that the middleware is functioning as expected.


validationparser/interceptor.go
```diff
@@ -0,0 +1,46 @@
...
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if v, ok := req.(validator); ok {
+			if err := v.Validate(); err != nil {
+				// Check what type of error we received
+				typeOfError := reflect.TypeOf(err)
+				valueOfError := reflect.ValueOf(err)
+				// If the error is a type of validation error then parse
+				if strings.Contains(typeOfError.String(), "ValidationError") && valueOfError.NumField() == 4 {
+					// Get the cause of error from the error struct
+					cause := valueOfError.FieldByName("Cause")
```
Calebjh: You should probably check that the error is a struct before doing Numfield() (even if the name contains ValidationError), https://stackoverflow.com/questions/42065226/how-to-determine-if-type-is-a-struct-in-golang
And also check that cause is a valid field https://golang.org/pkg/reflect/#Value.IsValid before doing TypeOf or ValueOf to prevent panics. 
I know I put you on this path, but I'd still love it if somebody had a better solution.

validationparser/interceptor.go
```diff
@@ -0,0 +1,46 @@
+package validationparser
```
dmacthedestroyer: not a great package name. Please refer to [package naming best practices](https://blog.golang.org/package-names)

validationparser/interceptor.go
```diff
@@ -0,0 +1,46 @@
...
+// user friendly error.
+func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if v, ok := req.(validator); ok {
+			if err := v.Validate(); err != nil {
+				// Check what type of error we received
+				typeOfError := reflect.TypeOf(err)
+				valueOfError := reflect.ValueOf(err)
+				// If the error is a type of validation error then parse
+				if strings.Contains(typeOfError.String(), "ValidationError") && valueOfError.NumField() == 4 {
```
dmacthedestroyer: this is a pretty tenuous assumption... Are you saying that it's a `lyft/validate` error if it ends with "ValidationError"?

validationparser/interceptor.go
```diff
@@ -0,0 +1,46 @@
...
+	Validate() error
+}
+
+// UnaryServerInterceptor returns grpc.UnaryServerInterceptor
+// that should be used as a middleware that parser through the protoc-gen-validate middleware and returns a
+// user friendly error.
+func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if v, ok := req.(validator); ok {
+			if err := v.Validate(); err != nil {
```
dmacthedestroyer: is this intended to replace the existing validation middleware we use? Otherwise, you're just calling validate again. A better approach would be some way of intercepting the response and checking for an error there

validationparser/interceptor.go
dmacthedestroyer: and add unit tests to verify all this!

validationparser/interceptor_test.go
```diff
@@ -0,0 +1,208 @@
...
+	return grpc.NewContextWithServerTransportStream(context.Background(), expectedStream)
+}
+
+// TestUnaryServerInterceptor_ValidationErrors will run mock validation errors to see if it parses correctly.
+func TestUnaryServerInterceptor_ValidationErrors(t *testing.T) {
+	ctx := DummyContextWithServerTransportStream()
+	tests := []struct {
+		name     string
+		want     mockRequestValidationError
+		expected error
```
dmacthedestroyer: `want` and `expected` are synonymous in the testing realm... what is meant by these?

My assumption is that `want` -> `input` and `expected` -> `want`

validationparser/interceptor.go
dmacthedestroyer: it's also kind of weird to own middleware that affects a specific open source project without it being in the open source project. We're basically "hacking" this project instead of adding value to the project we're relying on. Does that make any sense?

validationparser/interceptor.go
Calebjh: that should just entail doing 
```
v, err := handler(ctx, req)
>> error checking/alteration stuff...
return v, err
```
instead of calling Validate manually and working with the error it returns.

askurydzin:
I've looked through a code that is generated by validator middleware and it was really upsetting that they do not provide any methods to work with errors as interfaces (ValidationError with GetField, GetReason ecc). so the code implemented looks ok, however i cannot understand why do we need a separate interceptor implementation if we have errors interceptor that does exactly the same thing (if you want to catch validator errors you may want to plug errors middleware with your mapping function before it in a chain).

razamiDev:
@askurydzin we created it as a separate interceptor because we don't want our clients to be restricted with just our validator. Having it as a separate interceptors allows them to have more control on what they want to use and which validation interceptor.

askurydzin:
maybe providing a mapping function is enough for clients for not being restricted with our validator? i mean, we can just put this in a separate package inside errors as a preset, and if one wants to employ this mapping could just include it in our interceptor. This notion correlates with a preset for pq (postgres) errors, gorm errors and auth errors that are heterogeneous by nature but can be gathered under a single interface of mapping functions.

askurydzin:
ok, as per discussion with @razamiDev  for now we have no option to implement it in a consistent with, on one hand we have a standard interceptor that converts error to string and adds InvalidArgument gRPC code (making it very hard to parse the string back to errors container), on the other hand Raman's implementation that embeds error mapping. So for now I would say that everything looks ok to me, except few details.

validationparser/interceptor.go
```diff
@@ -0,0 +1,57 @@
...
+	valueOfError := reflect.ValueOf(err)
+
+	// Check if the error is a struct and a type of validation error that contains a cause.
+	if valueOfError.Kind() == reflect.Struct && strings.Contains(typeOfError.String(), "ValidationError") && !valueOfError.FieldByName("Cause").IsNil() {
+		cause := valueOfError.FieldByName("Cause")
+		typeOfCause := reflect.TypeOf(cause.Interface())
+		valueOfCause := reflect.ValueOf(cause.Interface())
+		if strings.Contains(typeOfCause.String(), "ValidationError") && valueOfCause.FieldByName("Field").String() != "" && valueOfCause.FieldByName("Reason").String() != "" {
+			// Retrieve the field and reason from the error
+			field := valueOfCause.FieldByName("Field").String()
```
askurydzin: maybe we should snake_case it?

validationparser/interceptor.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// UnaryServerInterceptor returns a new unary server interceptor that validates incoming messages.
+//
+// Invalid messages will be rejected with `InvalidArgument` before reaching any userspace handlers.
+func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if v, ok := req.(validator); ok {
+			if err := v.Validate(); err != nil {
+				return nil, err
```
askurydzin: return nil, ValidationHelper(err)

validationparser/interceptor.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+		}
+		return handler(ctx, req)
+	}
+}
+
+// MapValidationError function returns a MapFunc that checks if the error is of type validation
+func MapValidationError() errors.MapFunc {
```
askurydzin: I suggest to call it NewValidationErrorMapping()

validationparser/interceptor.go
razamiDev: We don't want to return the Validation Error here because the mapper only works with lyft validations. If there are other validations that exist we want to handle them differently we would want to add a mapper for each type of validation. That is why we want to just return the error so that our mapper can correctly handle it.

validationparser/interceptor.go
razamiDev: Could you describe to me the benefits of having a snake_case over the current camel case?

validationparser/interceptor.go
razamiDev: I updated it to ValidationErrorMapping because adding "New" might be a little confusing.

validationparser/interceptor.go
askurydzin: snake_case would correspond to an API field rather than it's internal representation in protobuf. So, user sends some_emails, and got 'invalid SomeEmails'

validationparser/interceptor.go
askurydzin: got it

errors/validationerrors/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```json
+{
+    "error": {
+        "status": 400,
+        "code": "INVALID_ARGUMENT",
+        "message": "Invalid PrimaryEmail: value must be a valid email address"
+    },
+    "fields": {
+        "PrimaryEmail": [
```
askurydzin: snake case?

razamiDev: Yes, I will update the README.

askurydzin:
LGTM, maybe regexp engine usage seems to me a bit overkill (sequential checks for IsLower/IsUpper would do it much faster), however performance for error converting is not of a big concern,

errors/mappers/validationerrors/interceptor.go
```diff
@@ -0,0 +1,105 @@
...
+	typeOfError := reflect.TypeOf(err)
+	valueOfError := reflect.ValueOf(err)
+
+	// Check if the error is a struct and a type of validation error that contains a cause.
+	if valueOfError.Kind() == reflect.Struct && strings.Contains(typeOfError.String(), "ValidationError") && !valueOfError.FieldByName("Cause").IsNil() {
+		cause := valueOfError.FieldByName("Cause")
+		typeOfCause := reflect.TypeOf(cause.Interface())
+
+		valueOfCause := reflect.ValueOf(cause.Interface())
+		if strings.Contains(typeOfCause.String(), "ValidationError") && valueOfCause.FieldByName("Field").String() != "" && valueOfCause.FieldByName("Reason").String() != "" && !valueOfCause.FieldByName("Cause").IsNil() {
```
ZachEddy: Might be nice to break this up into a few lines that are shorter because it extends pretty far off the page. I think [the 80 character limit](https://softwareengineering.stackexchange.com/questions/312889/how-to-follow-the-80-character-limit-best-practice-while-writing-source-code) is a good rule of thumb, but it's not something to obsess over.

PR #58, opened 2018-06-22
Add function to copy masked fields between two objects

Calebjh:
Was working on this a couple days ago as part of implementing PATCH for protoc-gen-gorm.

Will allow for a PATCH as:
- Read dbObject
- Call MergeWithMask(newObject, dbObject, field mask)
- Write dbObject

Could probably still use a few more test cases.


Evgeniy-L:
LGTM. I think this function is handy. Later we can consider generation of functions that can do the same thing without reflect.

Evgeniy-L:
@Calebjh, can you please take a look why CI build failed?

PR #59, opened 2018-06-26
Add proto-db field name conversion for collection operators

dkukharau:


gorm/utilities_test.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func TestHandleFieldPath(t *testing.T) {
+	tests := []struct {
+		fieldPath []string
+		dbName    string
+		err       bool
+	}{
+		{[]string{"name"}, "db_humans.name", false},
+		{[]string{"age"}, "db_humans.years", false},
```
Calebjh: Having `table_name.column` works for standard queries without an `AS` or join? I suppose that makes sense, but I don't know that I've seen it before.

gorm/collection_operators.go
```diff
@@ -77,7 +84,7 @@ func ApplyPagination(db *gorm.DB, p *query.Pagination) *gorm.DB {
 }
 
 // ApplyFieldSelection applies field selection operator fs to gorm instance db.
-func ApplyFieldSelection(db *gorm.DB, fs *query.FieldSelection) *gorm.DB {
+func ApplyFieldSelection(db *gorm.DB, fs *query.FieldSelection, obj interface{}) *gorm.DB {
```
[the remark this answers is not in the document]
dkukharau: Yeah, the work for field selection will be done after.

gorm/utilities_test.go
Evgeniy-L: If this works in Postgres, won't it be any problems with other DBs like MySQL.

dkukharau: For me It seems very natural to work. Table name is not required even in case of join and `AS` if there is no collision between column names across tables. I tested it on postgres and sqlite - works fine. And remember doing similar things with Oracle DB.

dkukharau:
@Calebjh Regarding invalid fields: `ApplyCollectionOperators` still allows passing unknown fields and doesn't raise an error in this case. This is done to support custom sorting/filtering, when e.g. some table is joined by an app dev, and toolkit knows nothing about it. I expect that validation of fields in collection operators will be handled in a separate middleware before they even come to gorm.

PR #60, opened 2018-06-27
Transfer collection operations via corresponding message fields

ystankevich:


gateway/fields.go
```diff
@@ -4,37 +4,16 @@ import (
...
 //retainFields function extracts the configuration for fields that
 //need to be ratained either from gRPC response or from original testRequest
 //(in case when gRPC side didn't set any preferences) and retains only
 //this fields on outgoing response (dynmap).
 func retainFields(ctx context.Context, req *http.Request, dynmap map[string]interface{}) {
-	fieldsStr, ok := Header(ctx, fieldsMetaKey)
-	if !ok && req != nil {
+	fieldsStr := ""
+	if req != nil {
 		//no fields in gprc response -> try to get from original testRequest
```
ystankevich: That retainFields is called by ForwardMessage where request is available as http.Request, so don't have access to message collection fields. As result, _fields will be parsed only from http.request

gateway/gateway.go
```diff
@@ -27,14 +31,74 @@ type gateway struct {
 	mux               *http.ServeMux
 }
 
+// CollectionOperationsInterceptor parse collection operators and stores in corresponding message fields
+func CollectionOperationsInterceptor(parentCtx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
```
ystankevich: That is client interceptor on  gateway, which parse  URL parameters and store in corresponding collection fields.

gateway/operator.go
```diff
@@ -5,152 +5,36 @@ import (
...
-
-	if v := vals.Get(limitQueryKey); v != "" {
-		mdmap[runtime.MetadataPrefix+limitMetaKey] = v
-	}
-
-	if v := vals.Get(pageTokenQueryKey); v != "" {
-		mdmap[runtime.MetadataPrefix+pageTokenMetaKey] = v
-	}
-
+	mdmap[QUERY_URL] = req.URL.String()
```
ystankevich: For gateway intercepter which parse collection ops, we need to have access to the URL parameters, so in MetadaAnnotator we just store it metadata

gateway/gateway.go
amaskalenka: It would better to have postfix `ClientUnaryInterceptor`, or probably move it to the `query` package at all .

ystankevich: CollectionOperationsInterceptor depends on non public variables like fieldsQueryKey

gateway/operator.go
```diff
@@ -5,152 +5,36 @@ import (
...
 	offsetQueryKey           = "_offset"
-	offsetMetaKey            = "operator-offset"
 	pageTokenQueryKey        = "_page_token"
-	pageTokenMetaKey         = "operator-page-token"
 	pageInfoSizeMetaKey      = "status-page-info-size"
 	pageInfoOffsetMetaKey    = "status-page-info-offset"
 	pageInfoPageTokenMetaKey = "status-page-info-page_token"
-	paginationContextKey     = "pagination-context"
+
+	QUERY_URL = "QUERY_URL"
```
Evgeniy-L: Do we really need this public?

gateway/middleware.go
```diff
@@ -101,7 +54,7 @@ func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
 	}
 }
 
-func setOp(req, op interface{}) error {
+func SetCollectionOps(req, op interface{}) error {
```
Evgeniy-L: I see that this method may return error like `testRequest value is not a struct`. I don't like `testRequest` here. It will be good to fix it and something more user friendly here (e.g. replace `testRequest` with `request`)

PR #61, opened 2018-06-27
Add transaction hook

friahi65:
- N* applications need to send notifications to another service about modified objects in a course of a request. We need to trigger the notification only when the transaction commits. 
- Adding a hook to toolkit transaction was an option that was discussed  in the team and in the toolkit scrum.
- The hook was added as part of transaction struct, it can be set by the applications.
- We need the context in the transaciton, so other middleware can put information about actual changes here. 
- Here is a link to an example on how this hook can be used in the application.
https://github.com/Infoblox-CTO/ddi.dns.config/compare/toolkit...friahi65:TK-45717-Design

### DIFF
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -41,14 +41,14 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
 // It works as a singleton to prevent an application of creating more than one
 // transaction instance per incoming request.
 type Transaction struct {
-	mu           sync.Mutex
-	parent       *gorm.DB
-	current      *gorm.DB
-	OnCommitHook func(ctx context.Context)
+	mu              sync.Mutex
+	parent          *gorm.DB
+	current         *gorm.DB
+	afterCommitHook []func(context.Context)
 }
 
-func (t *Transaction) SetOnCommitHook(f func(ctx context.Context)) {
-	t.OnCommitHook = f
+func (t *Transaction) AddAfterCommitHook(f []func(context.Context)) {
+	t.afterCommitHook = f
 }
 
 func BeginFromContext(ctx context.Context) (*gorm.DB, error) {
@@ -102,11 +102,12 @@ func (t *Transaction) Commit(ctx context.Context) error {
 	}
 	t.mu.Lock()
 	defer t.mu.Unlock()
-
 	t.current.Commit()
 	err := t.current.Error
-	if err == nil && t.OnCommitHook != nil {
-		t.OnCommitHook(ctx)
+	if err == nil && len(t.afterCommitHook) != 0 {
+		for i := range t.afterCommitHook {
+			t.afterCommitHook[i](ctx)
+		}
 	}
 	t.current = nil
 	return err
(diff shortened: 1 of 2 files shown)


gorm/transaction.go
```diff
@@ -41,9 +41,14 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
 // It works as a singleton to prevent an application of creating more than one
 // transaction instance per incoming request.
 type Transaction struct {
-	mu      sync.Mutex
-	parent  *gorm.DB
-	current *gorm.DB
+	mu           sync.Mutex
+	parent       *gorm.DB
+	current      *gorm.DB
+	OnCommitHook func(ctx context.Context)
```
dmacthedestroyer: let's make this unexportable (`onCommitHook`), and also make it a slice of `func(context.Context)` instead of a single hook. That way you can have multiple callbacks if needed

dmacthedestroyer: let's also call it `afterCommit` instead of `onCommit` to make it clear when this fires

gorm/transaction.go
```diff
@@ -41,9 +41,14 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
...
+}
+
+func (t *Transaction) SetOnCommitHook(f func(ctx context.Context)) {
```
dmacthedestroyer: following my previous comments, this would become `AddAfterCommitHook`

gorm/transaction.go
```diff
@@ -41,14 +41,14 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
...
-	OnCommitHook func(ctx context.Context)
+	mu              sync.Mutex
+	parent          *gorm.DB
+	current         *gorm.DB
+	afterCommitHook []func(context.Context)
 }
 
-func (t *Transaction) SetOnCommitHook(f func(ctx context.Context)) {
-	t.OnCommitHook = f
+func (t *Transaction) AddAfterCommitHook(f []func(context.Context)) {
```
dmacthedestroyer: the idea here would be that you add a single commit hook, and it appends it to the list:

```
func (t *Transaction) AddAfterCommitHook(hook func(context.Context)) {
    t.afterCommitHooks = append(t.afterCommitHooks, hook)
}
```

Better yet, you could make it variadic:
```
func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
    t.afterCommitHooks = append(t.afterCommitHooks, hooks...)
}
```

friahi65: you were perfectly clear btw, fault is all mine

gorm/transaction.go
```diff
@@ -91,15 +96,19 @@ func (t *Transaction) Rollback() error {
...
+func (t *Transaction) Commit(ctx context.Context) error {
 	if t.current == nil {
 		return nil
 	}
 	t.mu.Lock()
 	defer t.mu.Unlock()
-
 	t.current.Commit()
 	err := t.current.Error
+	if err == nil && len(t.afterCommitHook) != 0 {
```
dmacthedestroyer: nitpick: you don't need to check `len` because the `range` statement will work just fine on an empty slice

friahi65: true, I'll modify this.

PR #62, opened 2018-06-27
Add ability to add extra gateway options, e.g. marshaller/metadata

Calebjh:
Allows for extra marshaler options like extra metadata annotators or marshaler options for outputting zero values.

Naming can be a little messy/repetitive, e.g.
`gateway.WithGatewayOptions(runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{OrigName: true, EmitDefaults: true})),` 
or 
`gateway.WithGatewayOptions(runtime.WithMetadata(metadataFunc)),` 
suggestions are welcome.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -25,6 +25,7 @@ type gateway struct {
 	serverDialOptions []grpc.DialOption
 	endpoints         map[string][]registerFunc
 	mux               *http.ServeMux
+	gatewayMuxOptions []runtime.ServeMuxOption
 }
 
 // NewGateway creates a gRPC REST gateway with HTTP handlers that have been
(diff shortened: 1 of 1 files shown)


gateway/gateway.go
```diff
@@ -49,8 +50,8 @@ func NewGateway(options ...Option) (*http.ServeMux, error) {
 func (g gateway) registerEndpoints() (*http.ServeMux, error) {
 	for prefix, registers := range g.endpoints {
 		gwmux := runtime.NewServeMux(
-			runtime.WithProtoErrorHandler(ProtoMessageErrorHandler),
-			runtime.WithMetadata(MetadataAnnotator),
+			append(g.gatewayMuxOptions, runtime.WithProtoErrorHandler(ProtoMessageErrorHandler),
```
dmacthedestroyer: should our two default options override user-provided options, or vice-versa?

Calebjh: I was considering the same. I did it this way purely because it fit with the `append(slice, item, item)` that my brain prefers. Should I swap them?

PR #64, opened 2018-06-28
Add an option to require client TLS certificates and verification

### LINKED ISSUE #41: Fix TLS client certificate based authentication
Need to set up the server TLSConfig properly, including the ClientCAs and the tls.RequireAndVerifyClientCert

johnbelamaric:
This adds an additional TLS flag `-client-auth` which can take on these values:

 * `none` - the default, do not require client certificates: `tls.NoClientCert` [tls.ClientAuthType](https://golang.org/pkg/crypto/tls/#ClientAuthType)
 * `require` - require a cert but don't validate: `tls.RequireAnyClientCert`
 * `verify` - require and validate client cert: `tls. RequireAndVerifyClientCert`

Maybe `tls.VerifyClientCertIfGiven` would be better than the `require` one here. Or we can just allow them all.

Fixes #41


johnbelamaric:
This should be followed up with a `TLSClientFlags` struct in `flag.go` that provides a separate set of TLS options for use when acting as a client, rather than a server.

Evgeniy-L:
Closing as this is unlikely to be merged.

PR #65, opened 2018-06-29
Add metadata annotator and interceptor to enable field presence

Calebjh:
Spliced together some code from Dan's PR to grpc-gateway for field mask populating with metadata annotator and interceptor code from elsewhere in the toolkit to create a possible solution for field presence. Code and tests aren't very clean (yet), because my time and attention are mostly focused on another project. Behavior related to #58 

Internal discussion is still ongoing as to whether this constitutes a violation of proto3's design intentions.


Calebjh:
FieldMask is not _currently_ included in POST/PUT requests, this is true. However, it could be included, I see no significant cost to doing so. 

As far as consistency, I believe this matches our approach with collection operators (use an Annotator in the gateway to put them into the metadata, then an interceptor in the server to put them into the proto message).

The disadvantage to relying on the context values is that it is not friendly for grpc calls that don't go through the gateway. Using a FieldMask, you can specify which fields you are specifying versus leaving default, relying on the context would require the client to manually add metadata to the request to accomplish this, I believe. Much less intuitive.

Evgeniy-L:
@Calebjh, for collection operators we are now not putting them into metadata. We had two options:  metadata and define fields in proto -> this caused problems when you want to modify some collection operator in the code (when you use both options they get out of sync) -> decided to leave only option 2.

More details are in this [doc](https://docs.google.com/document/d/1nSgBRHaTp3fNcigACKNwZW80nmbd-KFo2bNMYuHyWeI/edit#) I believe.

In this case approach with FieldMask for POST/PUT/PATCH looks more consistent a believe.

Calebjh:
Hmm. In lieu of putting the collection operators into the context/metadata, we're now putting the URL query string into the metadata instead. I don't like the idea of putting the entire JSON body into the metadata just so that we can copy that pattern though. The idea is that putting it into the context is only for passing it from gateway -> server ( I believe that I saw registering the UnaryServerInterceptor as a dial option in the gateway is still a server-side interceptor), then it should be present in the proto, and should not be pulled from the context.

Calebjh:
Hmm. With further investigation, it looks like I may have been mistaken (confused by usage of the term "client" I think). I'll check if it's possible using gateway interceptors to populate the field mask without needing a client-side interceptor.

gateway/field_presence.go
```diff
@@ -0,0 +1,109 @@
...
+		} else if len(item.path) > 0 {
+			// otherwise, it's a leaf node so print its path
+			paths = append(paths, strings.Join(item.path, "."))
+		}
+	}
+
+	md := make(metadata.MD)
+	key := fieldPresenceMetaKey
+	for _, path := range paths {
+		md[key] = append(md[key], path)
```
dmacthedestroyer: what's this doing? it looks like you've got a slice of strings and your iterating the slice to make a slice of strings.

Could you just set the slice without iterating?

gateway/field_presence.go
```diff
@@ -0,0 +1,109 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/golang/protobuf/protoc-gen-go/generator"
+	"google.golang.org/genproto/protobuf/field_mask"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+const fieldPresenceMetaKey = "field-paths"
```
dmacthedestroyer: should this be some kind of object or something? I think it's best practice to avoid using anything that could collide as a key in the metadata. Or is this avoided somehow?

gateway/gateway.go
```diff
@@ -97,10 +98,13 @@ func ClientUnaryInterceptor(parentCtx context.Context, method string, req, reply
...
 	g := gateway{
-		serverAddress:     DefaultServerAddress,
-		endpoints:         make(map[string][]registerFunc),
-		serverDialOptions: []grpc.DialOption{grpc.WithInsecure(), grpc.WithUnaryInterceptor(ClientUnaryInterceptor)},
-		mux:               http.NewServeMux(),
+		serverAddress: DefaultServerAddress,
+		endpoints:     make(map[string][]registerFunc),
+		serverDialOptions: []grpc.DialOption{grpc.WithInsecure(), grpc.WithUnaryInterceptor(
+			grpc_middleware.ChainUnaryClient(
+				[]grpc.UnaryClientInterceptor{ClientUnaryInterceptor, PresenceClientInterceptor()}...),
```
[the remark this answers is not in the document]
Calebjh: That's fair. It shouldn't have to be added by default (it's just easier if it is, but I see your point), but that might mean adding another `WithOption` type command.

gateway/field_presence.go
Calebjh: It is a metadata key, not a context key, so it has to be a string (and hope/arrange for no collision). Context keys are arbitrary types, but the `metadata.MD` is a `map[string][]string`

gateway/field_presence.go
Calebjh: ... (_sheepishly_) yes, that's exactly what it's doing, and yes, it should just be directly set.

dmacthedestroyer: I'm glad I understood it, but I'm sad I didn't get to learn a new nifty Go trick :)

gateway/gateway.go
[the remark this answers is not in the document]
dmacthedestroyer: hmm, interesting. If this can't be done outside of some kind of internal usage then the server option needs to be expanded anyway, I suppose

gateway/field_presence.go
Calebjh: It's a low level obfuscatory trick. You're free to copy it if you'd like, but I'd advise against it.

PR #72, opened 2018-07-12
Add documentation to the auth package

ZachEddy:
# Add documentation to the auth package

This pull request adds documentation to the `auth` package that explains how to use the authorization middleware.

It covers interceptors at a basic level, then goes into more detail about the authorization interceptor and how to customize its behavior.

It's a WIP at the moment, but I wanted to open a PR to get feedback early-on in the process.

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -4,7 +4,7 @@ This package provides a server-side gRPC interceptor that interfaces with [Themi
 
 It is designed to give developers fine-grained control over who (e.g. specific users) or what (e.g. neighboring services) can access their gRPC service's business logic.
 
-## How it Works
+## Background
 
 If you're unfamiliar with gRPC interceptors and their intended use, please consider reading [this](https://github.com/grpc-ecosystem/go-grpc-middleware#middleware) brief explanation.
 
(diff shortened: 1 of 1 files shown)


auth/README.md
```diff
@@ -0,0 +1,133 @@
+# Authorization Interceptor
+
+This package provides a server-side gRPC interceptor that interfaces with Themis, a policy engine that is developed and maintained by Infoblox. It is designed to give developers fine-grained control over who (e.g. specific users) or what (e.g. neighboring services) can access their gRPC service's business logic.
```
dmacthedestroyer: maybe provide a link to the Themis repo?

auth/README.md
```diff
@@ -0,0 +1,133 @@
...
+```go
+// Define your grpc server
+myServer := grpc.NewServer(
+  rpc.StreamInterceptor(grpc_middleware.ChainStreamServer(
+    // Define your list of grpc interceptors
+  )),
+)
+```
+
+For more information about adding server-side gRPC interceptors, check out [the documentation](https://github.com/grpc-ecosystem/go-grpc-middleware) on the official gRPC GitHub repository.
```
dmacthedestroyer: I love your expository, and am happy to see it stay, but in all honesty I think this whole section could have been replaced with just this link :)

ZachEddy: Okay great! I'm definitely looking to trim this down as much as I can.

auth/README.md
```diff
@@ -0,0 +1,133 @@
...
+applicationID := "shopping-mall"
+// Define your grpc server
+myServer := grpc.NewServer(
+  grpc.UnaryInterceptor(grpc_middleware.ChainStreamServer(
+    // Include the authorization interceptor
+    auth.UnaryServerInterceptor(themisAddress, applicationID),
+  )),
+)
+```
+In the example above, the `themisAddress` variable corresponds to the host and port of Themis. The `applicationID`  maps the interceptor to a specific application. In a microservices environment, you might have multiple services that, as a unit, compose an application (e.g. a "petstore" service, a "coffee shop" service, and a "cellphone kiosk" service might belong to a "shopping mall" application). Each service has its own access control logic, but they're all under the same application umbrella.
```
dmacthedestroyer: is there any open-source info available on the web for this application id stuff? Is there any plan to?

auth/README.md
```diff
@@ -0,0 +1,133 @@
...
+authFunc := authorizer.AuthFunc()
+myServer := grpc.NewServer(
+  grpc.UnaryInterceptor(grpc_middleware.ChainStreamServer(
+    // include the authorization interceptor
+    grpc_auth.UnaryServerInterceptor(authFunc)
+  )),
+)
+```
+
+You'll likely find the Go documentation helpful, too.
```
dmacthedestroyer: ❤️

auth/README.md
ZachEddy: @markinos Thoughts on this? I added lots of detail because some developers haven't actually used gRPC interceptors before.

Should I just replace this section with a link?

auth/README.md
ZachEddy: No, I don't believe so. From what I know, the notion of an application identifier and how/where it would get used isn't well known 😕

John wrote-up a Google Doc about it a while ago, but I don't think it's progressed much since then.

dmacthedestroyer: this is kind of a puzzling feature, then, to have in an open source repository ;)

auth/README.md
dmacthedestroyer: My thought is that the intro would just be something like:

This package provides a server-side [gRPC interceptor](https://github.com/grpc-ecosystem/go-grpc-middleware) that interfaces with [Themis](https://github.com/infobloxopen/themis) to give developers fine-grained control over who (e.g. specific users) or what (e.g. neighboring services) can access their gRPC service's business logic.

dmacthedestroyer: or, even better: 

Behold! This package provides a server-side gRPC interceptor that interfaces with Themis to give developers fine-grained control over who (e.g. specific users) or what (e.g. neighboring services) can access their gRPC service's business logic.

auth/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+```sh
+godoc -http :6060
+```
+Visit [this link](http://localhost:6060/pkg/github.com/infobloxopen/atlas-app-toolkit/auth/) in your browser.
```
Evgeniy-L: `localhost:6060/pkg/` looks redundant

PR #74, opened 2018-07-13
Add toolkit support for integration testing

ZachEddy:
# Integration Testing

This pull request adds support for integration testing. There are a lot of additions, but the code itself isn't hugely complicated. Here's a summary of the package and each major file within it.

**The `intergration` package**

This package provides helpers that spin-up a basic testing environment to run integration tests repeatedly and reliably. The helpers are intended to be application-agnostic and can get reused by multiple teams.

In fact, some of these helpers aren't specific to integration testing, so maybe it's worth renaming this package to something more general like `test` instead (okay, maybe's that _too_ general).

**The `exec.go` file**

Integration testing might involve creating and running a Go binary, or maybe running a container to behave as a short-lived database. This file offers basic helpers that run command-line calls for these exact purposes.

It also provides a way to stop/destroy the binaries and containers that get created during testing.

**The `grpc.go` file**

To make gRPC calls, the request context needs to include a JWT for authorization. If it doesn't, then the client will get an error message. The `grpc.go` file has some basic helpers to build a standard request context.

**The `http.go` file**

This exists for the same reason `grpc.go` does: to make it easy to repeatedly send requests that might need specific headers/metadata.

**The `jwt.go` file**

This is concerned JWT-related functionality, like creating test tokens and getting a token with "standard" fields in the payload (e.g. an `AccountID` field).

**The `postgres.go` file**

This file is mostly responsible for spinning-up a Postgres container and exposing a set of functional options to configure the database. It also provides functionality to reset the database on a per-test basis.

### DIFF
--- a/integration/exec.go
+++ b/integration/exec.go
@@ -26,14 +26,13 @@ func RunBinary(binPath string, args ...string) (func(), error) {
 	return cancel, nil
 }
 
-// BuildSource builds a target Go package and gives the resulting binary
-// some user-defined name
+// BuildGoSource builds a target Go package and gives the resulting binary
+// some user-defined name. The function returned by BuildGoSource will remove
+// the binary that got created.
 func BuildGoSource(packagePath, output string) (func() error, error) {
 	cmdBuild := exec.Command("go", "build", "-o", output, packagePath)
 	if out, err := cmdBuild.CombinedOutput(); err != nil {
-		return nil, errors.New(fmt.Sprintf("unable to build package: %v (%s)",
-			err, string(out),
-		))
+		return nil, fmt.Errorf("unable to build package: %v (%s)", err, string(out))
 	}
 	return func() error {
 		return os.Remove(output)
@@ -51,18 +50,15 @@ func RunContainer(image string, dockerArgs, runtimeArgs []string) (func() error,
 	// launch the container
 	out, err := exec.Command(execDocker[0], execDocker[1:]...).CombinedOutput()
 	if err != nil {
-		return nil, errors.New(
-			fmt.Sprintf("%s: %s", err.Error(), string(out)),
-		)
+		return nil, fmt.Errorf("%s: %s", err.Error(), string(out))
 	}
 	// get the uuid from the container after it starts. this gets used to kill
 	// the container
-	var uuid string
-	if split := strings.Split(string(out), "\n"); len(split) == 0 {
+	split := strings.Split(string(out), "\n")
+	if len(split) == 0 {
 		return nil, errors.New("unable to get container uuid")
-	} else {
-		uuid = split[0]
 	}
+	uuid := split[0]
 	return func() error {
 		return exec.Command("docker", "kill", uuid).Run()
 	}, nil
--- a/integration/exec.go
+++ b/integration/exec.go
@@ -1,0 +1,69 @@
+package integration
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// RunBinary runs a target binary with a set of arguments provided by the user
+func RunBinary(binPath string, args ...string) (func(), error) {
+	abs, err := filepath.Abs(binPath)
+	if err != nil {
+		return nil, err
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	cmd := exec.CommandContext(ctx, abs, args...)
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+	return cancel, nil
+}
+
+// BuildSource builds a target Go package and gives the resulting binary
+// some user-defined name
+func BuildGoSource(packagePath, output string) (func() error, error) {
+	cmdBuild := exec.Command("go", "build", "-o", output, packagePath)
+	if out, err := cmdBuild.CombinedOutput(); err != nil {
+		return nil, errors.New(fmt.Sprintf("unable to build package: %v (%s)",
+			err, string(out),
+		))
+	}
+	return func() error {
+		return os.Remove(output)
+	}, nil
+}
+
+// RunContainer launches a detached docker container on the host machine. It
+// takes an image name, a list of "docker run" arguments, and a list of
+// arguments that get passed to the container runtime
+func RunContainer(image string, dockerArgs, runtimeArgs []string) (func() error, error) {
+	// build a list of args to pass to cmd.Exec
+	execDocker := append([]string{"docker", "run", "-d"}, dockerArgs...)
+	execRuntime := append([]string{image}, runtimeArgs...)
+	execDocker = append(execDocker, execRuntime...)
+	// launch the container
+	out, err := exec.Command(execDocker[0], execDocker[1:]...).CombinedOutput()
+	if err != nil {
+		return nil, errors.New(
+			fmt.Sprintf("%s: %s", err.Error(), string(out)),
+		)
+	}
+	// get the uuid from the container after it starts. this gets used to kill
+	// the container
+	var uuid string
+	if split := strings.Split(string(out), "\n"); len(split) == 0 {
+		return nil, errors.New("unable to get container uuid")
+	} else {
+		uuid = split[0]
+	}
+	return func() error {
+		return exec.Command("docker", "kill", uuid).Run()
+	}, nil
+}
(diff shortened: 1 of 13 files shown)


ZachEddy:
The build is failing because `go test` runs inside a build tool container. We can get around the issue by running this in the Travis build.

```
export BUILD_TYPE=local
```

While this would work, I'm personally in favor of removing the build tool entirely. Would this cause any issues?

integration/http.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// MakeStandardRequest issues an HTTP request a specific endpoint with Atlas-specific
+// request data (e.g. the authorization token)
+func MakeStandardRequest(method, url string, payload interface{}) (*http.Response, error) {
```
Evgeniy-L: For now I see this only used in tests. What concerns me here is that we will create a new client for every request. Maybe it makes sense to allow client reuse by passing a client as a parameter. This may be useful as well when client needs some configuration (e.g. not just standard http.Client{}).

Evgeniy-L: Or as an option function may return `http.Request` instead of `http.Response`

integration/jwt.go
```diff
@@ -0,0 +1,33 @@
+package integration
+
+import (
+	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/infobloxopen/atlas-app-toolkit/auth"
+)
+
+const (
+	// TestSecret dummy secret used for signing test JWTs
+	TestSecret = "some-secret-123"
```
Evgeniy-L: Do we need this public?

Evgeniy-L:
In terms of removing buildtool, I remember there was some reason for adding it but I cannot recall what exactly was it. We can always try and see what will happen :)

integration/http.go
ZachEddy: Good point. I see the problem with creating `http.Client` types over and over again. 

There's definitely room to improve. I like the idea of returning a `http.Request` type, so I'll start there.

integration/jwt.go
ZachEddy: Nope. Also, I need to comb through this code with `go vet` and `golint` and fix any issues that come up.

integration/grpc.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+	"fmt"
+
+	jwt "github.com/dgrijalva/jwt-go"
+	"google.golang.org/grpc/metadata"
+)
+
+// AppendTokenToOutgoingContext adds an authorization token to the gRPC
+// request context metadata. The user must provide a token field name like "token"
+// or "bearer" to this function. It is intended specifically for gRPC testing.
```
dmacthedestroyer: it would be good to add extra info here on example usage, or create a testable example demonstrating how to use this function

integration/postgres.go
```diff
@@ -0,0 +1,196 @@
...
+func WithVersion(version string) func(*testPostgresDB) {
+	return func(db *testPostgresDB) {
+		db.dbVersion = version
+	}
+}
+
+// WithMigrateFunction is used to rebuild the test Postgres database on a
+// per-test basis. Whenever the database is reset with the Reset() function, the
+// migrate function will rebuild the tables.
+func WithMigrateFunction(migrateFunction func(*sql.DB) error) func(*testPostgresDB) {
```
dmacthedestroyer: I was thinking about this more, and maybe it'd be nice to have a `WithUnMigrateFunction` or something that would tear down the database as well. The default setting can be your `DROP SCHEMA` command, but a user could, for instance, pass in the `golang-migrate` `migrate.down` command into it for finer control

integration/network.go
```diff
@@ -0,0 +1,47 @@
...
+
+type portError struct{ message string }
+
+// Error returns a string with the port-related error message
+func (p portError) Error() string {
+	return fmt.Sprintf("port discovery error: %s", p.message)
+}
+
+// GetOpenPortInRange finds the first unused port within specific range
+func GetOpenPortInRange(lowerBound, upperBound int) (int, error) {
```
dmacthedestroyer: it'd be great to have an even simpler function that takes no parameters and returns an int -- when I'm using this I usually get confused about what port range I should care about and then end up copying/pasting something from somewhere else. A default range would be nice to have

ZachEddy: Maybe something like this?

```
// GetOpenPort finds an unused port
func GetOpenPort() {
	GetOpenPortInRange(portRangeMin, portRangeMax)
}

// GetOpenPortInRange finds the first unused port within specific range
func GetOpenPortInRange(lowerBound, upperBound int) (int, error) {
...
}
```

dmacthedestroyer: I like the API, but don't like ignoring the error. Maybe either take in a `testing.T` or some `Fataller` interface that will let you call `.Fatal(interface{})`?

ZachEddy: > I like the API, but don't like ignoring the error.

Agreed. I left that out by accident.

> Maybe either take in a testing.T or some Fataller interface that will let you call .Fatal(interface{})?

Yeah, I was wondering about the whole `fatal` notion. I initially passed a `testing.T` to a number of these helpers, but removed the `testing.T` argument because IMO these helpers should not make the decision to fatal. The alternative is to call these helpers inside a `_test.go` file and have developers choose how they want to handle errors.

Whatever we decide, I think we should go "all or nothing" in favor of that decision. As in, all helpers should take a `testing.T` or none of them should. So, with that in mind, we should consider whether or not it makes sense for the other helpers to take a `testing.T` type.

integration/exec.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return func() error {
+		return os.Remove(output)
+	}, nil
+}
+
+// RunContainer launches a detached docker container on the host machine. It
+// takes an image name, a list of "docker run" arguments, and a list of
+// arguments that get passed to the container runtime
+func RunContainer(image string, dockerArgs, runtimeArgs []string) (func() error, error) {
```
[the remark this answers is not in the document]
ZachEddy: Nice idea! I'll take a look. I would definitely prefer using the Docker API because system commands feel clunky for some reason. Hopefully it'll make testing much easier, too.

integration/grpc.go
ZachEddy: Yeah, I like the idea of creating a testable example.

ZachEddy:
@Evgeniy-L Do you have any idea why the build is failing? I don't have any issue running `make test` locally. I'm out of ideas!

Evgeniy-L:
@ZachEddy, I have an idea that this may be related to latest changes in themis (using latest master will make our code unable to compile). Try merge/rebase latest master.

Evgeniy-L:
Make sure that Gopkg.toml points to specific revision for themis

integration/grpc_test.go
```diff
@@ -28,6 +29,28 @@ func TestAppendTokenToOutgoingContext(t *testing.T) {
 	}
 }
 
+func ExampleAppendTokenToOutgoingContext() {
+	// someFunc doesn't do anything, but in a real-world situation it might
+	// send a request to some grpc service that requires an authorization
+	// token
+	someFunc := func(context.Context) {
+		// send a request to some grpc service
```
dmacthedestroyer: you should print the contents of the token here to test that the context is properly decorated, then verify it with the `//Output:...` comment at the end of the func

integration/grpc_test.go
```diff
@@ -28,6 +29,28 @@ func TestAppendTokenToOutgoingContext(t *testing.T) {
...
+		},
+	).SignedString([]byte("some-secret"))
+	if err != nil {
+		log.Fatalf("unable to build token: %v", err)
+	}
+	ctxBearer := AppendTokenToOutgoingContext(
+		context.Background(), authToken, "Bearer",
+	)
+	someFunc(ctxBearer)
+}
```
dmacthedestroyer: does this test run with `go test`?  I think you'll need an `//Output:` comment in order for it to execute as a test

integration/postgres.go
```diff
@@ -60,17 +61,21 @@ func (db testPostgresDB) Reset() error {
...
-		"CREATE SCHEMA public;" +
-		"GRANT ALL ON SCHEMA public TO postgres;" +
-		"GRANT ALL ON SCHEMA public TO public;"
-	// drop all the tables in the test database
-	if _, err := dbSQL.Exec(resetQuery); err != nil {
-		return err
+	if db.migrateDownFunction != nil {
+		db.migrateDownFunction(dbSQL)
+	} else {
+		resetQuery := "DROP SCHEMA public CASCADE;" +
```
dmacthedestroyer: nitpick: you could instead have this be the default `db.migrateDownFunction` when calling `NewTestPostgresDB` and get rid of this `else` statement. Not a big deal, just pointing it out

PR #78, opened 2018-07-16
Make the JWT token field name configurable

ZachEddy:
# Configure JWT Token Field

~_This is a breaking change!_~

Currently, the authorization interceptor does not allow for someone to configure the field name of the authorization token. It assumes that all tokens come in the following format.

```
Authorization: token [insert some token]
```

This is problematic because the token field can be whatever the developer/organization wants.

```
Authorization: Bearer [insert some token]
```
```
Authorization: GrizzlyBear [insert some token]
```

This pull request gives developers a way to configure the name of the token field, but maintains a sensible default that allows the interceptor to be added easily to an application.

It's worth noting the documentation in #72 will need to change along with this request. I propose that we merge #72 and include documenation updates in this pull request.

### DIFF
--- a/auth/jwt.go
+++ b/auth/jwt.go
@@ -28,6 +28,9 @@ var (
 	errInvalidAssertion = errors.New("unable to assert value as jwt.MapClaims")
 )
 
+// GetJWTFieldWithTokenType gets the JWT from a context and returns the
+// specified field. The user must provide a token type, which prefixes the
+// token itself (e.g. "Bearer" or "token")
 func GetJWTFieldWithTokenType(ctx context.Context, tokenType, tokenField string, keyfunc jwt.Keyfunc) (string, error) {
 	token, err := getToken(ctx, tokenType, keyfunc)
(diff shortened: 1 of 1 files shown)


auth/jwt.go
```diff
@@ -25,33 +26,33 @@ var (
 )
 
 // GetJWTField gets the JWT from a context and returns the specified field
-func GetJWTField(ctx context.Context, field string, keyfunc jwt.Keyfunc) (string, error) {
-	token, err := getToken(ctx, keyfunc)
+func GetJWTField(ctx context.Context, tokenField, key string, keyfunc jwt.Keyfunc) (string, error) {
```
dmacthedestroyer: how would you feel about if, instead of breaking the contract, adding an overload (`GetJWTFieldWithTokenName` or something) that adds the functionality?

Calebjh:
Appears to be a valuable and important enhancement, but also a breaking change. I would like to see this pushed to next release for additional review time.

PR #84, opened 2018-07-24
field presence interceptor works for an arbitrarily-named field

dmacthedestroyer:
Change the behavior of the field presence interceptor to use reflection to find the first FieldMask regardless of what it's named


Calebjh:
There should be a corresponding change to the last section of the Readme for this

PR #87, opened 2018-07-27
Enable string comparison operators for filtering

dkukharau:


Makefile
```diff
@@ -38,4 +38,4 @@ check-fmt:
 	--go_out=plugins=grpc:. --grpc-gateway_out=logtostderr=true:. $(REPO)/server/testdata/test.proto
 
 .PHONY: gen
-gen: .gen-op .gen-errdetails .gen-errfields
+gen: .gen-query .gen-errdetails .gen-errfields .gen-servertestdata
```
Evgeniy-L: Do we really need `gen-servertestdata` here? I would prefer to exclude it from this target.

dkukharau: Up to you.

PR #91, opened 2018-08-01
Fix bug that can cause field masks to get mangled

Calebjh:
Sneaky bug is caused because underlying array of slice may or may not be deep copied during an append. This can cause later appends to overwrite the fields of earlier appends. Tests didn't detect because either structures weren't deep enough or didn't branch enough.

### DIFF
--- a/gateway/field_presence_test.go
+++ b/gateway/field_presence_test.go
@@ -20,6 +20,27 @@ func TestAnnotator(t *testing.T) {
 		`{}`: metadata.MD{},
 		`{`:  nil,
 		`{"one":{"two":"a", "three":[]}, "four": 5}`: {fieldPresenceMetaKey: []string{"Four", "One.Two", "One.Three"}},
+		`{
+  "name": "atlas",
+  "burden": {
+    "duration": "forever",
+    "weight": "earth",
+    "breaks": [],
+    "replacements": {
+			"hero": {
+	        "name": "hercules",
+	        "duration": "temporary",
+					"lineage": {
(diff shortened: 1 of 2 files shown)


Evgeniy-L:
How was it tested to be sure that it actually solves the problem?

Calebjh:
Ah, right. I discovered it in an internal project, and the variant of this function we have there. I added a test case in the private version, and tested it in the project code to make sure it resolved the issue. I could add another test case here, too, if you'd like.

dmacthedestroyer:
you should add the test case here.

gateway/field_presence_test.go
```diff
@@ -20,6 +20,27 @@ func TestAnnotator(t *testing.T) {
...
 		`{"one":{"two":"a", "three":[]}, "four": 5}`: {fieldPresenceMetaKey: []string{"Four", "One.Two", "One.Three"}},
+		`{
+  "name": "atlas",
+  "burden": {
+    "duration": "forever",
+    "weight": "earth",
+    "breaks": [],
+    "replacements": {
+			"hero": {
+	        "name": "hercules",
```
dmacthedestroyer: 💪

gateway/field_presence_test.go
```diff
@@ -20,6 +20,27 @@ func TestAnnotator(t *testing.T) {
...
+	        "duration": "temporary",
+					"lineage": {
+						"mother": "alcmena",
+						"father": "zeus"
```
dmacthedestroyer: ⚡️

PR #92, opened 2018-08-01
Change multi-tenant identifier from AccountID to account_id

ZachEddy:
# Change Multi-Tenant Identifier

This pull request changes the multi-tenant ID from `AccountID` to `account_id` for JWT-related purposes.

Also, I fixed a potential issue where `GetJWTFieldWithToken` would panic if it's given a `nil` context.


Calebjh:
Looks like this (seemingly small) change breaks the Jenkins test cases. I'm with Dan on this, I'd like to see support for both unless there's a compelling reason to disallow the former version.

dmacthedestroyer:
you can remove that `breaking change` label, I believe. Right?

PR #94, opened 2018-08-01
Add safety against paths leading to non-struct/non-primitive types

Calebjh:


gorm/fieldmask.go
```diff
@@ -24,6 +24,7 @@ func MergeWithMask(source, dest interface{}, mask *fieldmask.FieldMask) error {
 	if reflect.TypeOf(source) != reflect.TypeOf(dest) {
 		return errors.New("Types of source and destination objects do not match")
 	}
+pathsloop:
```
dmacthedestroyer: what kind of black magic is this?

gorm/fieldmask.go
```diff
@@ -36,6 +37,13 @@ func MergeWithMask(source, dest interface{}, mask *fieldmask.FieldMask) error {
 				dstVal = dstVal.Elem()
 				srcVal = srcVal.Elem()
 			}
+			// For safety, skip paths that will cause a panic to call FieldByName on
+			if dstVal.Kind() == reflect.Invalid || dstVal.Kind() == reflect.Array ||
```
dmacthedestroyer: did you get them all? Too bad there's no `CanFieldByName` like there is `CanAddr`, etc.

dmacthedestroyer: The [reflect.FieldByName](https://golang.org/pkg/reflect/#Value.FieldByName) documentation leads me to believe that this function will only work on `struct` kinds. Would this be a cleaner test?

gorm/fieldmask.go
Calebjh: https://www.ardanlabs.com/blog/2013/11/label-breaks-in-go.html

dmacthedestroyer: I'd say to use labels sparingly -- seeing a label in code sets off my abuse-of-control-flow alarms in general. I'm ok with it here, but please let's not do it everywhere :)

Calebjh: I guess I'll rethink my plan of refactoring everything to use labels and `goto`s instead of function calls then...

Really though, I think this is... either the first or second time I've ever used one. I'm quite sparing.

PR #95, opened 2018-08-03
Introduced dedicated interface to define resource name.

amaskalenka:
Added SetPlural flag option to produce plural form of resource name.


amaskalenka:
This fix is not backward compatible. 
The applications that rely on default behavior will be affected.
The applications that implement `XXX_MessageName` **will not be affected**

Now by default the singular form of resource name is returned in lower case format in comparison with previous version that returns plural form.

PR #96, opened 2018-08-03
Fix swagger definitions in query and rpc

askurydzin:
These changes will allow to treat our special atlas types in a bit different way (see https://github.com/infobloxopen/atlas-contacts-app/pull/65)


dmacthedestroyer:
This is really clean -- I like it! Does it actually work? Does it rely on modifications to protoc-gen-swagger?

askurydzin:
yes, sadly, but this strongly relies on these changes in protoc-gen-swagger, my patch forbids following messages that are declared as primitive types (string, number, ...)

dmacthedestroyer:
that's a pretty big catch :) Do you think this is possible? I really like this approach and think it'd be a nice addition to the swagger-codegen project as a whole

askurydzin:
https://github.com/infobloxopen/grpc-gateway/tree/master/protoc-gen-swagger/genswagger now these changes live here, so we can use our fork in gentool

PR #98, opened 2018-08-09
Enable gorm filtering by Identifier

dkukharau:
Breaking changes:
- ApplyCollectionOperators and other public function signatures are changed.

### DIFF
--- a/gorm/collection_operators.go
+++ b/gorm/collection_operators.go
@@ -1,21 +1,24 @@
 package gorm
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
-	"github.com/infobloxopen/atlas-app-toolkit/query"
+	"github.com/golang/protobuf/proto"
 	"github.com/jinzhu/gorm"
+
+	"github.com/infobloxopen/atlas-app-toolkit/query"
 )
 
 // ApplyCollectionOperators applies collection operators to gorm instance db.
-func ApplyCollectionOperators(db *gorm.DB, obj interface{}, f *query.Filtering, s *query.Sorting, p *query.Pagination, fs *query.FieldSelection) (*gorm.DB, error) {
-	db, fAssocToJoin, err := ApplyFiltering(db, f, obj)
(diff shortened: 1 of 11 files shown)


gorm/collection_operators.go
```diff
@@ -1,21 +1,24 @@
...
+	"github.com/golang/protobuf/proto"
 	"github.com/jinzhu/gorm"
+
+	"github.com/infobloxopen/atlas-app-toolkit/query"
 )
 
 // ApplyCollectionOperators applies collection operators to gorm instance db.
-func ApplyCollectionOperators(db *gorm.DB, obj interface{}, f *query.Filtering, s *query.Sorting, p *query.Pagination, fs *query.FieldSelection) (*gorm.DB, error) {
-	db, fAssocToJoin, err := ApplyFiltering(db, f, obj)
+func ApplyCollectionOperators(ctx context.Context, db *gorm.DB, obj interface{}, pb proto.Message, f *query.Filtering, s *query.Sorting, p *query.Pagination, fs *query.FieldSelection) (*gorm.DB, error) {
```
chinmayb: Is it not a breaking change ? we recently removed `ctx` from here because its anti pattern

gorm/utilities.go
```diff
@@ -15,7 +18,7 @@ import (
 // according to obj GORM model. If fieldPath cannot be found in obj then original fieldPath is returned
 // to allow tables joined by a third party.
 // If association join is required to resolve the field path then it's name is returned as a second return value.
-func HandleFieldPath(fieldPath []string, obj interface{}) (string, string, error) {
+func HandleFieldPath(ctx context.Context, fieldPath []string, obj interface{}) (string, string, error) {
```
Evgeniy-L: why do we need context here?

gorm/utilities.go
```diff
@@ -6,7 +6,10 @@ import (
 	"reflect"
 	"strings"
 
+	"context"
```
Evgeniy-L: please check fmt and organize imports properly.

gorm/joins.go
```diff
@@ -10,7 +11,7 @@ import (
 // - association table name
 // - source join keys
 // - target join keys
-func JoinInfo(obj interface{}, assoc string) (string, []string, []string, error) {
+func JoinInfo(ctx context.Context, obj interface{}, assoc string) (string, []string, []string, error) {
```
Evgeniy-L: why we need context here?

PR #99, opened 2018-08-10
Convert atlas.rpc.Identifier to snake case instead on lowercase

Evgeniy-L:


util/snake.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"regexp"
+	"strings"
+)
+
+var matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
+var matchAllCap = regexp.MustCompile("([a-z0-9])([A-Z])")
+
+// CamelToSnake takes a camelcase string and returns a snake case string.
+func CamelToSnake(str string) string {
```
Evgeniy-L: Moved from  errors/mappers/validationerrors/interceptor.go

PR #105, opened 2018-08-17
Auth remove interceptor

ZachEddy:
# Remove Authorization Interceptor

The authorization interceptor has been moved to a dedicated middleware respository internally. This pull request removes the interceptor, but keeps some of the JWT-related functionality that can be used in applications.

This shouldn't get merged immediately.

Hey @dmacthedestroyer, check out all of these deletions!


dmacthedestroyer:
:feelsgood: 

Maybe instead of deleting all this stuff, we add a [deprecated flag](https://stackoverflow.com/a/36360323/437456) for a while. Then, when we're confident this won't just nuke a thousand downstream repositories, we can delete it.

Evgeniy-L:
This might be a good idea to live with deprecated flag for a while.
BTW will it nuke out contacts-app?

ZachEddy:
> This might be a good idea to live with the deprecated flag for a while.

Agreed, but are some dependency managemented-related issues with Themis that prevent the authorization interceptor updates from getting merged internally. I might be able to fix the issue by moving some variables around to prevent `dep` from including any files that use Themis. 

> BTW will it nuke out contacts-app?

Yep. I just opened this pull request ahead of time so we can show people the exact changes that we'll be making.

PR #109, opened 2018-08-23
 Utility functions to support page token handling

Evgeniy-L:
- Changes for ErrorContainer to return error message
- Utility functions to support page token handling (moved from contacts app)


errors/container.go
```diff
@@ -49,7 +49,7 @@ func InitContainer() *Container {
 }
 
 // Error function returns "Unknown".
```
amaskalenka: The comment should be changed as well.

PR #112, opened 2018-09-03
Add logic to retain fields in array and change appropriate test

EvgenyMilev:


gateway/fields_test.go
```diff
@@ -104,18 +104,99 @@ func TestRetainSingleResult(t *testing.T) {
...
+func TestRetainArray(t *testing.T) {
+	data := `
+	{
+		"result": [
+		  {
+			"x": "1",
+			"y": [
+				{
+				 "a": "2", 
+                 "b": "3"
```
Evgeniy-L: please check formatting as it looks inconsistent

gateway/fields_test.go
```diff
@@ -104,18 +104,99 @@ func TestRetainSingleResult(t *testing.T) {
...
+		   "b":[
+			  {"c":"ccc",
+			   "d":"ddd",
+			   "x":"xxx"
+		      },
+			  {"c":"ccc2",
+			   "d":"ddd2",
+			   "x":"xxx2"
+		      }
+			],
```
Evgeniy-L: I see that you changed all test for nested object to arrays. With this we do not have tests for nested objects. Please revert this and add some new tests for use cases with arrays.

PR #113, opened 2018-09-06
 Fix bug - Identifier content is not validated

EvgenyMilev:


Evgeniy-L:
why this PR contains changes from #112?

PR #119, opened 2018-10-10
Removed reflect usage from validation errors and updated tests.

razamiDev:


Evgeniy-L:
Is this correct that app will have to update lyft to newer version once they switch to new version of toolkit?

razamiDev:
> Is this correct that app will have to update lyft to newer version once they switch to new version of toolkit?

@Evgeniy-L yes that is correct.

PR #121, opened 2018-10-15
Add interceptor which populates request-scoped logger with account_id field

dkukharau:


auth/interceptor.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus/ctxlogrus"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+)
+
+// UnaryServerInterceptor returns grpc.UnaryServerInterceptor which populates request-scoped logrus logger with account_id field
+func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
```
dmacthedestroyer: this name meets the convention, but in this case it doesn't make a lot of sense...  What would you think about a more specific name? `AccountIDLoggerInterceptor`? Or something?

dkukharau:
Initially I named it `auth.LogrusUnaryServerInterceptor`. I didn't want to tie with account_id because there are some other fields which we may want to add to logger from JWT(e.g. user_id) later. Then I realized that this interceptor can also be used for JWT signature validation, so I chose current name. Now I'm thinking that validation should be a separate middleware(since somebody may not use logrus, but still want JWT validation), so the initial name seems ok.

auth/interceptor.go
dkukharau: Initially I named it auth.LogrusUnaryServerInterceptor. I didn't want to tie with account_id because there are some other fields which we may want to add to logger from JWT(e.g. user_id) later. Then I realized that this interceptor can also be used for JWT signature validation, so I chose current name. Now I'm thinking that validation should be a separate middleware(since somebody may not use logrus, but still want JWT validation), so the initial name seems ok.

PR #122, opened 2018-10-17
Add per-request context-logger level and pseudo-tracing tags

Calebjh:
An annotator and server (unary) interceptor for custom per-request log-level and a `log-trace-key` field for searching logs.

Could probably use additional documentation and some test cases...

### DIFF
--- a/logging/annotator.go
+++ b/logging/annotator.go
@@ -1,0 +1,34 @@
+package logging
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"google.golang.org/grpc/metadata"
+)
+
+// HTTP Headers
+const logLevelHeaderKey = "log-level"
+const logFlagHeaderKey = "log-trace-key"
+
+// Metadata keys used to pass from gateway to service
+const logLevelMetaKey = "log-level"
+const logFlagMetaKey = "log-trace-key"
+
+// Name of field to be logged if included
+const logFlagFieldName = "log-trace-key"
+
+// Annotator is a function that reads the http headers of incoming requests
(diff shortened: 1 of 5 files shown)


logging/annotator.go
```diff
@@ -0,0 +1,34 @@
...
+// HTTP Headers
+const logLevelHeaderKey = "log-level"
+const logFlagHeaderKey = "log-trace-key"
+
+// Metadata keys used to pass from gateway to service
+const logLevelMetaKey = "log-level"
+const logFlagMetaKey = "log-trace-key"
+
+// Name of field to be logged if included
+const logFlagFieldName = "log-trace-key"
```
Evgeniy-L: This seems similar to having a custom "Request-Id" (e.g. if we can set Request-Id from outside). It is possible to specify request-id from the outside if you use request-id interceptor. Are there any other use cases for this?

logging/interceptor.go
```diff
@@ -0,0 +1,58 @@
...
+		logFlag, hasFlag := gateway.Header(ctx, logFlagMetaKey)
+		newLogger := CopyLoggerWithLevel(entry.Logger, lvl)
+		fields := entry.Data
+		if hasFlag {
+			fields[logFlagFieldName] = logFlag
+		}
+		newCtx := ctxlogrus.ToContext(ctx, newLogger.WithFields(fields))
+		res, err = handler(newCtx, req)
+
+		// propagate any new or or changed fields from later interceptors back up
```
Evgeniy-L: `or` is written twice

logging/annotator.go
Calebjh: Request-ID should always be unique, this isn't necessarily unique.

PR #124, opened 2018-10-19
Fix marshal-unmarshal-marshal hack in response forwarder

dkukharau:
`jsonpb` marshaler should only be used for marshaling `protobuf` messages.
In our ResponseForwarder hack, when we marshal-unmarshal-marshal protobuf message, we should use standard json marshaler for unmarshaling-marshaling, because don't deal with proto at this stage. That will fix the problem related to map nil values.
P.S. This hack, which probably causes a significant performance penalty, should be removed eventually when we're ready to commit a new version of API syntax because there will be no need of custom response body manipulations anymore.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -27,7 +27,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:5a8d32b3e55e592f903fead865fe933164388ef3748f23430f9ca54a0dbd87cc"
+  digest = "1:b0add830aeeb55d56127cb60199eca80cbff045ee0cd1ed4b82153fd17288b5b"
   name = "github.com/golang/protobuf"
   packages = [
     "jsonpb",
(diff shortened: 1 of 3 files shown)


Gopkg.lock
```diff
@@ -10,24 +10,24 @@
...
 [[projects]]
-  digest = "1:ffe9824d294da03b391f44e1ae8281281b4afc1bdaa9588c9097785e3af10cec"
+  digest = "1:a2c1d0e43bd3baaa071d1b9ed72c27d78169b2b269f71c105ac4ba34b1be4a39"
   name = "github.com/davecgh/go-spew"
   packages = ["spew"]
   pruneopts = "UT"
-  revision = "8991bc29aa16c548c550c7ff78260e27b9ab7c73"
-  version = "v1.1.1"
+  revision = "346938d642f2ec3594ed81d874461961cd0faa76"
+  version = "v1.1.0"
```
Evgeniy-L: I see that we have versions decreased for multiple plugins. Do we have a reason for it?

dkukharau: `dep` did that for some reason. Just updated vendors.

dkukharau: `dep` did that for some reason. Just updated vendors.

dkukharau: `dep` did that for some reason. Just updated vendors.

PR #129, opened 2018-10-26
 Readme for health package

Evgeniy-L:


health/README.md
```diff
@@ -0,0 +1,29 @@
+# Health Package
+
+Package contains tools that allow to add REST endpoint for handling `health` and `readiness` requests.
+Package also provides implementation for most typical checks that may be done to figure our if service is healthy/ready.
+
+## REST Handler
```
ZachEddy: It's not really a REST API, so much as just a plain HTTP Endpoint that serves one specific purpose. IMO, I think you can just say _HTTP Handler_ or _Probe Endpoints_. If you're looking for synonyms, maybe check out the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/).

health/README.md
```diff
@@ -0,0 +1,29 @@
...
+	)
+```
+
+In a sample above `readinessCheck` and `livenessCheck` are functions that perform a single check.
+It is possible to add multiple checks independently and handler will call them in sequence.
+
+## Checks
+
+Packange provides several predefined checks:
+ - `DNSProbeCheck(host string, timeout time.Duration)` returns a Check that determines whether a service with specified dns name is reachable or not using net.Resolver's LookupHost method.
```
ZachEddy: "dns" should be uppercase.

health/README.md
```diff
@@ -0,0 +1,29 @@
...
+ - `HTTPGetCheck(url string, timeout time.Duration)` returns a Check that performs an HTTP GET request to the specified URL. It fails if timeout was exceeded or non-200-OK status code returned.
```
ZachEddy: Might help to change Check to `Check` to denote that it's a `struct` inside the package.

health/README.md
```diff
@@ -0,0 +1,29 @@
+# Health Package
```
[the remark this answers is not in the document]
Evgeniy-L: I picked this name as it is a most common pattern. We already have `RequestID Package`, `Authorization/Authentication Package`, `Gateway Package`, etc. Lets stick to this pattern.

PR #130, opened 2018-10-26
Update README file for gateway package

ZachEddy:
# Gateway README Refactor

This pull request updates the README file for the `gateway` package. I added a bit of extra documentation to the README, but I mostly just changed things around a little bit.

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -1,13 +1,15 @@
-# Gateway Package
+# Gateway
 
-### Resources and Collections
+This package contains helper functions that support creating, configuring, and running a gRPC REST gateway that is REST syntax compliant. Google already provides a lot of documentation related to the gRPC gateway, so this README will mostly serve to link to existing docs.
 
-#### How to define REST API Endpoints in my proto scheme?
+> gRPC is great — it generates API clients and server stubs in many programming languages, it is fast, easy-to-use, bandwidth-efficient and its design is combat-proven by Google. However, you might still want to provide a traditional RESTful API as well. Reasons can range from maintaining backwards-compatibility, supporting languages or clients not well supported by gRPC to simply maintaining the aesthetics and tooling involved with a RESTful architecture.
 
-You can map your gRPC service methods to one or more REST API endpoints.
-See [this reference](https://cloud.google.com/service-management/reference/rpc/google.api#http) how to do it.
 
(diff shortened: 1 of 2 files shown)


gateway/README.md
```diff
@@ -105,9 +104,10 @@ func (s *myServiceImpl) MyMethod(ctx context.Context, req *MyRequest) (*MyRespon
 }
 ```
 
-If you do not use any custom outgoing header matcher you would see something like that:
+If you do not use any custom outgoing header matcher, you will see something like this.
+
 ```sh
-> curl -i http://localhost:8080/contacts/v1/contacts
+$ curl -i http://localhost:8080/contacts/v1/contacts
```
Evgeniy-L: Just a random thought. In contacts app we currently have `http://localhost:8080/v1/contacts`, but here it doesn't even matter for this example. I was thinking if it makes sense to change it to something neutral like `http://localhost:8080/data`. Up to you...

gateway/README.md
```diff
@@ -118,55 +118,46 @@ Content-Length: 2
...
+	"github.com/infobloxopen/atlas-app-toolkit/gateway"
 )
 
 func init() {
 	forward_App_ListObjects_0 = gateway.ForwardResponseMessage
 }
 ```
 
-You can also refer [example app](https://github.com/github.com/infobloxopen/atlas-contacts-app/pb/contacts/contacts.overwrite.pb.gw.go)
+You can also refer [example app](https://github.com/infobloxopen/atlas-contacts-app/blob/master/pkg/pb/contacts.overwrite.pb.gw.go).
```
Evgeniy-L: This app is outdated but we can still refer it at least for this example. My concern is that some developers may consider using this app as an example. Do you think we should still refer old contacts app?

gateway/README.md
```diff
@@ -118,55 +118,46 @@ Content-Length: 2
...
 
-#### Which forwarders I need to use to comply our REST API?
+#### Conforming to REST API Syntax
```
Evgeniy-L: `REST API Syntax` phrase may be confusing for someone from outside Infoblox. Not sure if we should be concerned about it :)

gateway/README.md
ZachEddy: Agreed. I'll remove this sentence.

PR #134, opened 2018-10-31
Changes to request-id package readme

Calebjh:


requestid/README.md
```diff
@@ -19,11 +24,21 @@ func main() {
       grpc.UnaryInterceptor(
         grpc_middleware.ChainUnaryServer(  // middleware chain
           ...
+          [grpc_logrus.UnaryServerInterceptor(logrus.NewEntry(logger)),]
```
Evgeniy-L: Why `grpc_logrus` is in square brackets?

Calebjh: I've seen square brackets used some places for optional arguments, and the logging interceptor isn't strictly required for the Request-ID so I wanted to show somehow that this is where it should go, but it's "optional". I could drop them if you'd prefer though.

Evgeniy-L: Lets leave it the way it is at the moment

PR #135, opened 2018-11-02
Logging package readme expansion

Calebjh:
Fill out the readme a little more, match others betters

### DIFF
--- a/logging/README.md
+++ b/logging/README.md
@@ -1,13 +1,79 @@
-## Per Request Logger Settings
+# Logging
 
-The functions here support request scoped log settings.
-Specifically, the log level and a custom log field can be set with the ctxlogrus context logger based on http Header values with the grpc-gateway or via grpc metadata.
+This package supports extended settings for request scoped logging.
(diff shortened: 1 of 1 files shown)


logging/README.md
```diff
@@ -1,13 +1,79 @@
...
+}
+```
+
+For grpc-gateway support, the `MetadataAnnotator` should also be added to the gateway.
+Using the toolkit's server package, that setup looks something like this:
+```golang
+import (
+  ...
+  "github.com/infobloxopen/atlas-app-toolkit/logging"
+	"github.com/infobloxopen/atlas-app-toolkit/server"
```
razamiDev: Is there a reason this is spaced more than the line above it ?

logging/README.md
```diff
@@ -1,13 +1,79 @@
...
+func main() {
+  gatewayOptions := []grpcruntime.ServeMuxOption{
+    grpcruntime.WithMetadata(logging.MetadataAnnotator),
+    ...
+  }
+
+  server.NewServer(
+    server.WithGrpcServer(grpcServer),
+    server.WithGateway(
+			gateway.WithGatewayOptions(gatewayOptions...),
```
razamiDev: Same here with extra spacing

PR #139, opened 2018-11-06
add script and instruction for protoc plugins debug

burov:
I provide some kind of tutorial and utill scripts for protoc plugins debugging

### DIFF
--- a/tools/debugging/Dockerfile
+++ b/tools/debugging/Dockerfile
@@ -1,0 +1,13 @@
+FROM golang:1.10.0 AS builder
+
+LABEL stage=server-intermediate
+
+WORKDIR /go/src/github.com/infobloxopen/protoc-gen-atlas-validate
+COPY . .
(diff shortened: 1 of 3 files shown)


Evgeniy-L:
I remember we discussed if this need to go to toolkit, [gentool](https://github.com/infobloxopen/atlas-gentool) or a separate package. I think we had idea of having it in [gentool](https://github.com/infobloxopen/atlas-gentool)

burov:
@dkukharau We want review from you

tools/debugging/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.10.0 AS builder
```
dkukharau: Why have this file here if we've already put it in README?

burov: Because we use this file in ```docker build``` command, you must change it and use as a default behavior

PR #141, opened 2018-11-12
Remove page-info handling from gateway

dkukharau:


gateway/fields.go
```diff
@@ -25,15 +25,17 @@ func retainFields(ctx context.Context, req *http.Request, dynmap map[string]inte
 
 	fields := query.ParseFieldSelection(fieldsStr)
 	if fields != nil {
-		for _, result := range dynmap {
-			if results, ok := result.([]interface{}); ok {
-				for _, r := range results {
-					if m, ok := r.(map[string]interface{}); ok {
-						doRetainFields(m, fields.Fields)
+		for k, result := range dynmap {
+			if k == "result" || k == "results" {
```
Evgeniy-L: Per our syntax responses like below are valid. But in this case ratainFields logic will not be executed.
```
{
  "success": {
    "status": 200, 
    "message": "Read 360 items", 
    "code": "OK"
  },
  "rpz_hits": [{...}, {...}, ...]
}
```
We can check for `page` instead, but not sure that this is optimal solution. What do you think is a best way to address this?

dkukharau: I think a check for `page` is the only alternative. It will work as a temporary solution.
Anyway, I consider this field retaining functionality to go away from the gateway eventually since it's very harmful from performance perspective.

Evgeniy-L: Per current REST Syntax:
```
In any scenario where the response is to send back paging parameters (as listed in the table below), these should be contained in a page field in the top-level of the response object.
```

Evgeniy-L: We may discuss options during the day tomorrow

PR #142, opened 2018-11-14
Add helper function that can prevent errors and reduce log noise

Calebjh:
Readme update should describe situation well. Internal dev shows that when the leading underscore is not used for collection operators in the URL query strings, things break, and when they are included grpc logging always logs that it doesn't know what to do with it. By including a function like this, that's fixed! (Should create an internal version that includes other query strings that we might commonly use that are specially typed)

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -29,7 +29,7 @@ message GetMessageRequest {
   string user_id = 2;
 }
 ```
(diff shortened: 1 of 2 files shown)


Calebjh:
Hmmm. I suppose that's a good point. The reason I chose response.go was because that's where the other function `gateway.ForwardResponseMessage`, that also has to be manually overridden for each RPC, lives.

PR #144, opened 2018-11-16
Add request info

burov:
Add metadata annotator that extract some information about request and save it to context, work in progress need readme

### DIFF
--- a/requestinfo/README.md
+++ b/requestinfo/README.md
@@ -1,7 +1,6 @@
 # Request Info
 
-Request info contains information about request 
-
+Request Info contains following information about the request:
 ```
    ApplicationName string
    ResourceType    string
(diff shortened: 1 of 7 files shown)


requestinfo/README.md
```diff
@@ -0,0 +1,32 @@
+# Request Info
+
+This package provide annotator witch extract information from request and return it as metadata  
+## Adding support for Request Info
+
+You can enable support for Request Info in your gRPC-Server by adding the annotator to the middleware chain.
+However, the **ordering is important**.
+If you also use the grpc_logrus interceptor, the request-id middleware should be later in the middleware chain, but should also be before any other service middlewares to ensure it is present in the context to be included in those requests.
```
Evgeniy-L: I don't think we should mention `request-id middleware` here :)

requestinfo/README.md
```diff
@@ -0,0 +1,32 @@
...
+}
+```
+
+## Extracting the Request Info
+
+Once the middleware is included, the following function
+```golang
+requestInfo, err := requestinfo.FromContext(ctx)
+```
+can extract the request info anywhere it is needed.
```
Evgeniy-L: I think we should also mention the content of request info somewhere in a readme

PR #146, opened 2018-11-22
Remove apply paging to empty paging

burov:
Remove apply paging to empty paging

### DIFF
--- a/gorm/collection_operators.go
+++ b/gorm/collection_operators.go
@@ -36,7 +36,6 @@ func ApplyCollectionOperators(ctx context.Context, db *gorm.DB, obj interface{},
 
 	db = ApplyPagination(ctx, db, p)
 
-
 	db, err = ApplyFieldSelection(ctx, db, fs, obj)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


gorm/collection_operators.go
```diff
@@ -34,7 +34,9 @@ func ApplyCollectionOperators(ctx context.Context, db *gorm.DB, obj interface{},
 		return nil, err
 	}
 
-	db = ApplyPagination(ctx, db, p)
+	if !IsPaginationEmpty(p) {
```
dkukharau: Does it fix the issue? If offset is specified and limit is not, then the query will still return 1000 records. I think we agreed to entirely disable default limit.

gorm/collection_operators.go
```diff
@@ -123,3 +125,15 @@ func ApplyFieldSelection(ctx context.Context, db *gorm.DB, fs *query.FieldSelect
 	}
 	return db, nil
 }
+
+func IsPaginationEmpty(p *query.Pagination) bool {
+	if p == nil {
+		return true
+	}
+
+	if p.PageToken == "" && p.Limit == 0 && p.Offset == 0 {
```
dkukharau: `gorm` package shouldn't care of PageToken since it doesn't support it.

PR #150, opened 2019-01-08
Change hardcoded fallback error to match new syntax

Calebjh:


gateway/errors.go
```diff
@@ -87,7 +87,7 @@ func (h *ProtoErrorHandler) StreamHandler(ctx context.Context, headerWritten boo
 }
 
 func (h *ProtoErrorHandler) writeError(ctx context.Context, headerWritten bool, marshaler runtime.Marshaler, rw http.ResponseWriter, err error) {
-	const fallback = `{"code":"INTERNAL","status":500,"message":"%s"}`
+	const fallback = `{"error":[{"message":"%s"}]}`
```
johnsdouglass: What happens if the message supplied by Sprintf on line 139 has double quotes in it? Is that possible? Looks to me like the JSON would then be invalid.

Admittedly, this would have been a problem in the previous commit.

PR #151, opened 2019-01-18
Enable quote(single/double) escaping in filtering expressions

dkukharau:
Duplicate single/double quotes to use them in filtering expressions.

### DIFF
--- a/query/filtering_lexer_test.go
+++ b/query/filtering_lexer_test.go
@@ -7,7 +7,7 @@ import (
 )
 
 func TestFilteringLexer(t *testing.T) {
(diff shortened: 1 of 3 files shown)


query/filtering_lexer_test.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func TestFilteringLexer(t *testing.T) {
-	lexer := NewFilteringLexer(`()14 13.23 'abc'"bcd" field1 and or  not == eq ne != match ~ nomatch !~ gt > ge >= lt < le <= null := ieq [1,5, 6] ['Hello','World'] in `)
+	lexer := NewFilteringLexer(`()14 13.23 'abc'"bcd" field1 and or  not == eq ne != match ~ nomatch !~ gt > ge >= lt < le <= null := ieq [1,5, 6] ['Hello','World'] in '''""' """''"`)
```
Evgeniy-L: we should document this in README as current escaping logic is not obvious.

dkukharau: I would expect it's documented in REST API Syntax.

PR #153, opened 2019-02-18
Discard gRPC metadata from gateway response

dkukharau:
This information is useless for REST clients.

### DIFF
--- a/gateway/header.go
+++ b/gateway/header.go
@@ -80,12 +80,9 @@ func HeaderN(ctx context.Context, key string, n int) (val []string, found bool) 
 	}
 }
 
-// PrefixOutgoingHeaderMatcher prefixes outgoing gRPC metadata with
-// runtime.MetadataHeaderPrefix ("Grpc-Metadata-").
-// It behaves like the default gRPC-Gateway outgoing header matcher
-// (if none is provided as an option).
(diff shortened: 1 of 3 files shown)


gateway/header.go
```diff
@@ -80,12 +80,9 @@ func HeaderN(ctx context.Context, key string, n int) (val []string, found bool)
...
 }
 
-// PrefixOutgoingHeaderMatcher prefixes outgoing gRPC metadata with
-// runtime.MetadataHeaderPrefix ("Grpc-Metadata-").
-// It behaves like the default gRPC-Gateway outgoing header matcher
-// (if none is provided as an option).
+// PrefixOutgoingHeaderMatcher discards all grpc header metadata.
 func PrefixOutgoingHeaderMatcher(key string) (string, bool) {
-	return fmt.Sprintf("%s%s", runtime.MetadataHeaderPrefix, key), true
+	return "", false
```
Evgeniy-L: Looks like this way we completely remove an ability to pass data from grpc layer to user. For reported bug this seems to work fine, but how we are going to handle cases when this is actually required?

dkukharau: This is the default OutgoingHeaderMatcher. One is able to implement its own and apply to the gateway.

PR #156, opened 2019-03-07
Add code-to-level for gateway logging interceptor

Calebjh:
Initial behavior had gateway logging interceptor firing if the request failed before making it to the server, but always logging at the Info level despite the error. This should improve that handling by using the same logic as the standard server logging interceptor.

### DIFF
--- a/logging/gateway_interceptor.go
+++ b/logging/gateway_interceptor.go
@@ -22,6 +22,7 @@ type gwLogCfg struct {
 	noRequestID   bool
 	acctIDKeyfunc jwt.Keyfunc
 	withAcctID    bool
+	codeToLevel   grpc_logrus.CodeToLevel
 }
 
 // GWLogOption is a type of function that alters a gwLogCfg in the instantiation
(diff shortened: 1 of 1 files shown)


logging/gateway_interceptor.go
```diff
@@ -61,15 +62,32 @@ func EnableAccountID(o *gwLogCfg) {
...
+		o.codeToLevel = codeFunc
+	}
+}
+
+type sentinelKeyType struct{}
 
-const sentinelKey = sentinelKeyType(0)
+var sentinelKey = sentinelKeyType{}
+
+func SentinelValueFromCtx(ctx context.Context) (value, ok bool) {
```
Evgeniy-L: I do not think we have use case when someone will need this call. In current implementation `sentinelKey` introduces coupling between `GatewayLoggingInterceptor` and `GatewayLoggingInterceptor`. I would not encourage additional coupling with other interceptors. Do you mind removing this at least until we will face a use case when this is required?

Calebjh: This was actually already a direct ask from an internal consumer, otherwise I wouldn't have included it. For the exact same reason I did the coupling, they had something they want to log, but didn't want to log in both the server and gateway.

PR #157, opened 2019-03-15
add support for JSONB conditions in postgres

burov:
Known issues:
  * JSONB on nested fields not implemented yet

### DIFF
--- a/gorm/utilities.go
+++ b/gorm/utilities.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/protoc-gen-go/generator"
 	jgorm "github.com/jinzhu/gorm"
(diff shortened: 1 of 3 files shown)


gorm/utilities.go
```diff
@@ -40,6 +41,62 @@ func HandleFieldPath(ctx context.Context, fieldPath []string, obj interface{}) (
...
+	return fmt.Sprintf("%s %s '{%s}'", dbPath, operator, strings.Join(fieldPath[1:], ",")), "", nil
+}
+
+func isRawJSON(values ...string) bool {
+	if len(values) == 0 {
+		return false
+	}
+
+	for _, v := range values {
+		if !strings.HasPrefix(v, "{") || !strings.HasSuffix(v, "}") {
```
Evgeniy-L: consider move to regexp checking. Also think if cases with leading/trailing spaces are valid

PR #160, opened 2019-04-05
add check before start database

burov:


dkukharau:
Not sure what is this supposed to solve. Could you describe the use case?

burov:
@dkukharau First of all I add name to prevent idle database containers running and you might know you can't have two containers with the same name in the same time to prevent this I add check  before start container, and if we have container with the same name we are waiting until this name become free. This  prevent CI failing when two PRs send in close time frames

dkukharau:
@burov What do you mean by idle container? The name is randomly selected by the docker engine if it wasn't specified. Moreover, I don't see a problem in running concurrent postgres containers(given they're attached to different host ports). I would better try to make CI jobs independent rather than trying to synchronize between them.

burov:
@dkukharau But if panic happens, container will  running while you don't stop it manually, on Jenkins node.

dkukharau:
@burov The container will stay running in either way: if you specified the name explicitly or it was chosen randomly. In any case, you should have a proper cleanup in your tests, no matter if it fails or not.

burov:
@dkukharau  Yep but only one, and we will see that in output of the CI and delete it.

dkukharau:
If you care about the Jenkins node so much and want to track your containers I would still suggest to properly setup your tests cleanup. But if don't want to do this, another acceptable option for me would to add an option to specify the container name, and generate that name with some common prefix in your tests. This way you'll be able to track your containers, manually cleanup them(probably periodically) and still keep your jobs independent. I definitely vote against this single name pattern to be common for all atlas apps.

PR #162, opened 2019-04-25
Disconnect collection operators ToGorm logic from protobuf

Evgeniy-L:
This is "Work in progress" and is intended to demonstrate the proposal for removing dependency on protobuf for ColectionOperators -> Gorm conversion

We may also consider creating some abstraction layer on top of `paging` and `field selection` similar to what you see here for `filtering` and `sorting`


gorm/converter.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/protoc-gen-go/generator"
+
+	"github.com/infobloxopen/atlas-app-toolkit/query"
+	"github.com/infobloxopen/atlas-app-toolkit/rpc/resource"
+)
+
+type DefaultPbToOrmConverter struct {
+	processor FilteringConditionProcessor
```
burov: Shall we use composition there ? Maybe it better to implement all interfaces in separated structs and after, embed them into DefaultPbToOrmConverter like we did in CollectionOperatorsConverter?

Evgeniy-L:
Marking as `breaking changes` as some of the public functions may become unavailable.

PR #164, opened 2019-05-03
check in vendors so project builds

drewwells:
The project does not build on the master branch of `github.com/dgrijalva/jwt-go`. I found that the vendor folder was not checked in. Vendoring or Modules are required to make `go get` work. This checks in the vendor folder so the project can build.

https://golang.org/cmd/go/#hdr-Vendor_Directories

```
[~/src/github.com/infobloxopen/atlas-app-toolkit/auth] [master]
-> % go build
# github.com/infobloxopen/atlas-app-toolkit/auth
./jwt.go:41:28: invalid type assertion: token.Claims.(jwt.MapClaims) (non-interface type map[string]interface {} on left)
./jwt.go:80:12: undefined: jwt.Parser
./jwt.go:88:52: undefined: jwt.MapClaims
```


Evgeniy-L:
I do not really like the idea of having `vendor` folder in toolkit repo. Toolkit do not provide any executable and acts as a storage for common functions which may be reused by various apps.

I remember this was discussed long time ago when we faced a problem with nested vendor directories for apps using glide. At that point we agreed to commit vendor for applications and do not commit vendor for non-application repos.

PR #165, opened 2019-05-07
Err fields

Calebjh:
Currently, it isn't possible to return an error with any additional fields (using WithError can have extra fields, but is in addition to any error returned by the RPC). This change adds a function that can be called when returning an error to also specify extra details, as well as a function which toggles automatic inclusion of the `status` and `code` fields in response statuses.

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -169,11 +169,49 @@ func (s *myService) MyMethod(req *MyRequest) (*MyResponse, error) {
 ### Response Format
 Unless another format is specified in the request `Accept` header that the service supports, services render resources in responses in JSON format by default.
 
-Services must embed their response in a Success JSON structure.
+By default for a successful RPC call only the proto response is rendered as JSON, however for a failed call a special format is used, and by calling special methods the response can include additional metadata.
 
-The Success JSON structure provides a uniform structure for expressing normal responses using a structure similar to the Error JSON structure used to render errors. The structure provides an enumerated set of codes and associated HTTP statuses (see Errors below) along with a message.
+The `WithSuccess(ctx context.Context, msg MessageWithFields)` function allows you to add a `success` block to the returned JSON.
+By default this block only contains a message field, however arbitrary key-value pairs can also be included.
+This is included at top level, alongside the assumed `result` or `results` field.
 
-The Success JSON structure has the following format. The results tag is optional and appears when the response contains one or more resources.
+Ex.
+```json
+{
+  "success": {
+    "foo": "bar",
+    "baz": 1,
+    "message": <message-text>
+  },
+  "results": <service-response>
+}
+```
+
+The `WithError(ctx context.Context, err error)` function allows you to add an extra `error` to the `errors` list in the returned JSON.
+The `NewWithFields(message string, kvpairs ...interface{})` function can be used to create this error, which then includes additional fields in the error, otherwise only the error message will be included.
+This is included at top level, alongside the assumed `result` or `results` field if the call succeeded despite the error, or alone otherwise.
+
+Ex.
+```json
+{
+  "errors": [
+    {
+      "foo": "bar",
+      "baz": 1,
+      "message": <message-text>
+    }
+  ],
+  "results": <service-response>
+}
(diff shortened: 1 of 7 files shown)


gateway/errors.go
```diff
@@ -223,14 +267,34 @@ func WithSuccess(ctx context.Context, msg MessageWithFields) {
 	grpc.SetTrailer(ctx, md)
 }
 
-func errorsAndSuccessFromContext(ctx context.Context) (errors []map[string]interface{}, success map[string]interface{}) {
+func WithCodedSuccess(ctx context.Context, c codes.Code, msg string, args ...interface{}) {
```
razamiDev: Add a comment here for WithCodedSuccess

gateway/status_test.go
```diff
@@ -15,22 +15,28 @@ import (
 
 func TestStatus(t *testing.T) {
 	// test REST status from gRPC one
-	stat := HTTPStatus(context.Background(), status.New(codes.OK, "success message"))
+	stat, statName := HTTPStatus(context.Background(), status.New(codes.OK, "success message"))
```
razamiDev: Could you please add a couple more tests.

gateway/response.go
```diff
@@ -36,8 +36,14 @@ var (
 	ForwardResponseMessage = NewForwardResponseMessage(PrefixOutgoingHeaderMatcher, ProtoMessageErrorHandler, ProtoStreamErrorHandler)
 	// ForwardResponseStream is default implementation of ForwardResponseStreamFunc
 	ForwardResponseStream = NewForwardResponseStream(PrefixOutgoingHeaderMatcher, ProtoMessageErrorHandler, ProtoStreamErrorHandler)
+
+	setStatusDetails = false
 )
 
+func IncludeStatusDetails(withDetails bool) {
```
razamiDev: Add a comment here as well.

gateway/errors.go
```diff
@@ -119,16 +125,26 @@ func (h *ProtoErrorHandler) writeError(ctx context.Context, headerWritten bool,
...
+		restErr["status"] = statusStr
+	}
 
-	errs, _ := errorsAndSuccessFromContext(ctx)
+	errs, _, overrideErr := errorsAndSuccessFromContext(ctx)
 	restResp := &RestErrs{
 		Error: errs,
 	}
-	restResp.Error = append(restResp.Error, restErr)
+	logrus.Infof("%v", overrideErr)
```
razamiDev: I see you are adding a log here... 
I don't think we have added logs in the other parts of the tool. I think we should maybe add a little more info if we are going to include it. So that a developer can know where it is coming from.

Calebjh: Whoops. I probably left that in while I was doing manual testing...

gateway/response_test.go
```diff
@@ -97,11 +97,11 @@ func TestForwardResponseMessage(t *testing.T) {
 		t.Fatalf("failed to unmarshal JSON response: %s", err)
 	}
 
-	if v.Success["code"] != CodeName(Created) {
+	if v.Success["status"] != CodeName(Created) {
 		t.Errorf("invalid status code: %s - expected: %s", v.Success["code"], CodeName(Created))
 	}
 
-	if v.Success["status"].(float64) != http.StatusCreated {
+	if v.Success["code"].(float64) != http.StatusCreated {
```
razamiDev: Double check this. Look at the line below as it is giving error for v.Success["status"] and is checking v.Success["code"]

gateway/README.md
```diff
@@ -169,27 +169,61 @@ func (s *myService) MyMethod(req *MyRequest) (*MyResponse, error) {
 ### Response Format
 Unless another format is specified in the request `Accept` header that the service supports, services render resources in responses in JSON format by default.
 
-Services must embed their response in a Success JSON structure.
+By default for a successful RPC call only the proto response is rendered as JSON, however for a failed call a special format is used, and by calling special methods the response can include additional metadata.
```
razamiDev: Thanks for adding this! It will be really helpful.

PR #166, opened 2019-05-29
Add request-id annotator

chinmayb:
Currently if `Request-ID` is passed during the REST call in headers is not reflected in the logs. The request grpc middleware looks for `Request-ID` in `ctx`. But it can not find it even though its passed explicitly. This helps in debugging.

This PR  creates an metadata annotator that appends the request-Id in `ctx` in grpc gateway.

Demo:
 Before:
```
 curl  -X GET -H "Request-Id: test-request" "http://localhost:8080/xyz/v1/abc"
 respose:
  {"error":[{"message":"Internal error occurred. For more details see log for request 65680fc0-e59e-49d0-93f6-dcb9c13771db"}]
```
 Now:
```
 curl  -X GET -H "Request-Id: test-request" "http://localhost:8080/xyz/v1/abc"
 respose:
  {"error":[{"message":"Internal error occurred. For more details see log for request test-request "}]
```

### DIFF
--- a/requestid/metadata_annotator.go
+++ b/requestid/metadata_annotator.go
@@ -1,0 +1,20 @@
+package requestid
+
+import (
+	"context"
+	"net/http"
+
+	"google.golang.org/grpc/metadata"
+)
+
+func NewRequestIDAnnotator() func(context.Context, *http.Request) metadata.MD {
+	return func(ctx context.Context, req *http.Request) metadata.MD {
+		if req.Header.Get(DefaultRequestIDKey) != "" {
+			md := make(metadata.MD)
+			md.Set(DefaultRequestIDKey, req.Header.Get(DefaultRequestIDKey))
+			return md
+		} else {
+			return nil
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


requestid/metadata_annotator.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+	"net/http"
+
+	"google.golang.org/grpc/metadata"
+)
+
+func NewRequestIDAnnotator() func(context.Context, *http.Request) metadata.MD {
+	return func(ctx context.Context, req *http.Request) metadata.MD {
+		if req.Header.Get(DefaultRequestIDKey) != "" {
```
chinmayb: Do we need to check for `X-Request-ID` as well here?  Because by defualt Nginx Doesnt log `Request-ID` now. It logs only `X-Request-ID`

kd7lxl:
Is this required only because `Request-ID` is not in the header matcher? AuthN solved the problem like this: https://github.com/Infoblox-CTO/athena-authn-svc/blob/master/cmd/server/main.go#L45-L57

chinmayb:
@kd7lxl As you suggested moving the header matcher func to atlas-cli

PR #168, opened 2019-05-31
CustomIncomingHeaderMatcher func is used to add custom headers

chinmayb:
to be matched from incoming http requests, If this returns true
the header will be added to grpc context.

E.g. Apps can add headers like `Request-ID` to be accepted from REST
requests.

### DIFF
--- a/gateway/header.go
+++ b/gateway/header.go
@@ -114,9 +114,10 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
 	}
 }
 
-// CustomIncomingHeaderMatcher func is used to add custom headers to be matched
-// from incoming http requests, If this returns true the header will be added to grpc context
-func CustomIncomingHeaderMatcher(headerNames ...string) func(string) (string, bool) {
+// ExtendedDefaultHeaderMatcher func is used to add custom headers to be matched
+// from incoming http requests, If this returns true the header will be added to grpc context.
+// This function also passes through all the headers that runtime.DefaultHeaderMatcher handles.
+func ExtendedDefaultHeaderMatcher(headerNames ...string) func(string) (string, bool) {
 	customHeaders := map[string]bool{}
 	for _, name := range headerNames {
 		customHeaders[strings.ToLower(name)] = true
--- a/gateway/header.go
+++ b/gateway/header.go
@@ -113,3 +113,19 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
 		}
 	}
 }
+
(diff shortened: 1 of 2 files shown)


gateway/header_test.go
```diff
@@ -62,3 +62,52 @@ func TestPrefixOutgoingHeaderMatcher(t *testing.T) {
...
+
+func TestCustomHeaderMatcher(t *testing.T) {
+	var customMatcherTests = []struct {
+		name          string
+		customHeaders []string
+		in            string
+		isValid       bool
+	}{
+		{
+			name:          "Customer headers | success",
```
kd7lxl:
```suggestion
			name:          "Custom headers | success",
```

gateway/header_test.go
```diff
@@ -62,3 +62,52 @@ func TestPrefixOutgoingHeaderMatcher(t *testing.T) {
...
+			customHeaders: []string{"Request-ID", "ophid"},
+			in:            "Request-Id",
+			isValid:       true,
+		},
+		{
+			name:          "Customer headers| failure",
```
kd7lxl:
```suggestion
			name:          "Custom headers| failure",
```

gateway/header.go
```diff
@@ -113,3 +113,19 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
 		}
 	}
 }
+
+// CustomIncomingHeaderMatcher func is used to add custom headers to be matched
+// from incoming http requests, If this returns true the header will be added to grpc context
+func CustomIncomingHeaderMatcher(headerNames ...string) func(string) (string, bool) {
```
kd7lxl: Not sure I like the name of this. It could be used for either incoming or outgoing headers, so `Incoming` can be removed. And nothing about the name implies that it's additive with the default header matcher. The description is good, maybe name it like it is described:
```suggestion
func AdditionalHeaderMatcher(headerNames ...string) func(string) (string, bool) {
```

gateway/header.go
```diff
@@ -113,3 +113,20 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
 		}
 	}
 }
+
+// ExtendedDefaultHeaderMatcher func is used to add custom headers to be matched
+// from incoming http requests, If this returns true the header will be added to grpc context.
+// This function also passes through all the headers that runtime.DefaultHeaderMatcher handles.
+func ExtendedDefaultHeaderMatcher(headerNames ...string) func(string) (string, bool) {
```
kd7lxl: 👍 I like this name.

PR #177, opened 2019-10-16
Fix formating and add new versions of Golang to CI

burov:
Fix format issue which does not allow us to run Golang 1.12 + version on CI


.travis.yml
```diff
@@ -6,6 +6,9 @@ services:
 go:
   - "1.9"
   - "1.10"
```
drewwells: can we drop 1.9 and 1.10?

gateway/field_presence_test.go
```diff
@@ -17,14 +17,14 @@ import (
...
 func TestAnnotator(t *testing.T) {
-	for input, expect := range map[string]metadata.MD{
-		``:   metadata.MD{fieldPresenceMetaKey: nil},
-		`{}`: metadata.MD{fieldPresenceMetaKey: nil},
-		`{`:  nil,
-		`{"objects":[{"one": {"two":"a", "three":[]}, "four": 5}, {"one":{"two":"a", "three":[]}, "four": 5}]}`: {fieldPresenceMetaKey: []string{"Four$One.Two$One.Three", "Four$One.Two$One.Three"}},
-		`{"one":{"two":"a", "three":[]}, "four": 5}`:                                                            {fieldPresenceMetaKey: []string{"Four$One.Two$One.Three"}},
-		`{"one": {}}`:                                                                                           {fieldPresenceMetaKey: []string{"One"}},
-		`{
+	tests := map[string]metadata.MD{}
```
drewwells: Did this output also fail?

#go 1.12.5
```
gofmt -s -w gateway/field_presence_test.go
```

```
-> % git diff --patch
diff --git a/gateway/field_presence_test.go b/gateway/field_presence_test.go
index 96de41d..92e7521 100644
--- a/gateway/field_presence_test.go
+++ b/gateway/field_presence_test.go
@@ -18,12 +18,12 @@ import (

 func TestAnnotator(t *testing.T) {
        for input, expect := range map[string]metadata.MD{
-               ``:   metadata.MD{fieldPresenceMetaKey: nil},
-               `{}`: metadata.MD{fieldPresenceMetaKey: nil},
+               ``:   {fieldPresenceMetaKey: nil},
+               `{}`: {fieldPresenceMetaKey: nil},
                `{`:  nil,
                `{"objects":[{"one": {"two":"a", "three":[]}, "four": 5}, {"one":{"two":"a", "three":[]}, "four": 5}]}`: {fieldPresenceMetaKey: []string{"Four$One.Two$One.Three", "Four$One.Two$One.Three"}},
                `{"one":{"two":"a", "three":[]}, "four": 5}`:                                                            {fieldPresenceMetaKey: []string{"Four$One.Two$One.Three"}},
-               `{"one": {}}`:                                                                                           {fieldPresenceMetaKey: []string{"One"}},
+               `{"one": {}}`: {fieldPresenceMetaKey: []string{"One"}},
                `{
   "name": "atlas",
   "burden": {
```

burov: Yes, gofmt tools in Golang 1.10(I tested it with 1.10.8) and Golang 1.12(I tested it with 1.12.12) format this code in different ways

.travis.yml
burov: I don't think we can delete them now, because it can bring some errors if anybody will use functions from new versions of Golang in atlas-app-toolkit. Now we don't have some specific requirements for Golang version due to this we keep old versions for  a while, and gonna think how to deprecate supporting of old Golang version In future

PR #178, opened 2019-11-04
Add wrapper for using middleware

rdv821:


server/server.go
```diff
@@ -67,6 +71,14 @@ func NewServer(opts ...Option) (*Server, error) {
 	}
 	s.HTTPServer.Handler = mux
 
+	var handler http.Handler
+	if s.middleware != nil {
+		handler = s.middleware(mux)
+	} else {
+		handler = mux
+	}
+	s.HTTPServer.Handler = handler
```
dkukharau: It can probably be simplified:

```
	if s.middleware != nil {
		s.HTTPServer.Handler = s.middleware(s.HTTPServer.Handler)
	}
```

server/server.go
```diff
@@ -31,6 +31,7 @@ type Server struct {
 	initializers      []InitializerFunc
 	initializeTimeout time.Duration
 	registrars        []func(mux *http.ServeMux) error
+	middleware        Middleware
```
burov: From my point of view field "middleware" should be repeated to allow use to define chain of middleware, solution with one middleware from my point of view isn't enough flexible

server/server.go
burov: As I mentioned in comment above, we should consider defining multiple middleware not only one. 

``` go
    for _, m := range s.middleware {
         s.HTTPServer.Handler = m(s.HTTPServer.Handler)
    }
```

dkukharau:
As discussed with @burov, it's useful to add support for chaining middlewares. The preferable approach would probably be to add a utility function to be called on the client side, similar to [grpc_middleware.ChainUnaryServer](https://godoc.org/github.com/grpc-ecosystem/go-grpc-middleware#ChainUnaryServer).
```go
func ChainMiddlewares(middlewares ...Middleware) Middleware {
...
}
```
@rdv821 Could you add to your PR?

server/server.go
```diff
@@ -31,6 +31,7 @@ type Server struct {
 	initializers      []InitializerFunc
 	initializeTimeout time.Duration
 	registrars        []func(mux *http.ServeMux) error
+	middleware        []Middleware
```
dkukharau: `middlewares`

server/server.go
```diff
@@ -67,6 +71,10 @@ func NewServer(opts ...Option) (*Server, error) {
 	}
 	s.HTTPServer.Handler = mux
 
+	for _, m := range s.middleware {
```
dkukharau: You should probably wrap them backwards, so that the order of chain execution matches the order they are provided by the client.

server/server_test.go
```diff
@@ -127,6 +127,44 @@ func TestWithHandler(t *testing.T) {
...
+
+	go s.Serve(nil, httpL)
+	defer s.Stop()
+
+	resp, err := http.Get(fmt.Sprint("http://", httpL.Addr().String(), "/v1/hello?name=test"))
+	if resp.Header.Get("status") != "ok" {
+		t.Error("Expected another status")
+	}
+
+	resp2, err := http.Get(fmt.Sprint("http://", httpL.Addr().String(), "/v1/hello?name=test"))
```
dkukharau: Probably it doesn't make sense to make the same request twice.

server/server.go
```diff
@@ -129,6 +137,22 @@ func WithGateway(options ...gateway.Option) Option {
...
+func WithMiddlewares(middleware ...Middleware) Option {
+	return func(s *Server) error {
+		s.middlewares = append(s.middlewares, middleware...)
+		s.Sort()
+		return nil
+	}
+}
+
+// Sort args for middleware
+func (s *Server) Sort() {
```
dkukharau: Shouldn't be a public method.

server/server.go
```diff
@@ -129,6 +137,22 @@ func WithGateway(options ...gateway.Option) Option {
...
+	for i, j := 0, len(s.middlewares)-1; i < j; i, j = i+1, j-1 {
+		s.middlewares[i], s.middlewares[j] = s.middlewares[j], s.middlewares[i]
```
dkukharau: Could you also update the tests to check that the resulting order is right?

server/server_test.go
```diff
@@ -127,6 +127,51 @@ func TestWithHandler(t *testing.T) {
...
+}
+
+func testParamVerify(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("param", "status")
+		h.ServeHTTP(w, r)
+	})
+}
+
+func newTestParamVerify(t *testing.T, h http.Handler) http.Handler {
```
Evgeniy-L: This seems not be be used anywhere.

dkukharau:
@Evgeniy-L @burov any more feedback from your side?

PR #179, opened 2019-11-15
first pass of resource id

drewwells:
```
Application services require a mechanism to encode the identity of a particular resource across applications. The resource identifiers consist of typed guid containing a version, resource type, region and unique id.

In database tables, the resource identifier should be stored in its original form. The typed guid will be used in inter-service contracts. It may be beneficial to store the typed guid in other formats but should be done in addition to the full representation. This identifier is globally unique and sufficient for use as a Primary Key. 

What a typed guid looks like:

    blox1.<RESOURCE TYPE>.[REGION].<UNIQUE ID>

blox1.account..3kj4bhjf3h4fiuokn3idgfdhgfo2kjknlk2ns34iu6iub6bs
blox1.host.us-east1.asygyiu32y2u3y4b5buboiufuy2fo34uy2cytdtfdg

```

### DIFF
--- a/bloxid/interface.go
+++ b/bloxid/interface.go
@@ -1,0 +1,17 @@
+package bloxid
+
+// ID implements the interface for parsing a resource identifier
+type ID interface {
+	// String returns the complete resource ID
+	String() string
+	// ShortID returns a shortened ID that will be locally unique
+	ShortID() string
+	// Version returns a serialized representation of the ID version
+	// ie. `V0`
+	Version() string // V0
+	// Type returns entity type ie. `host`
+	Type() string
+	// Region is optional and returns the cloud region that
+	// the resource is found in ie. `us-east-1`
+	Region() string
+}
--- a/bloxid/rand.go
+++ b/bloxid/rand.go
@@ -1,0 +1,22 @@
+package bloxid
+
+import (
+	"crypto/rand"
+)
+
+const (
+	DefaultEntropySize = 40
+)
+
+func randDefault() []byte {
+	return randBytes(DefaultEntropySize)
+}
+
+func randBytes(size int) []byte {
+	b := make([]byte, size/2)
+	_, err := rand.Read(b)
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
--- a/bloxid/v0.go
+++ b/bloxid/v0.go
@@ -1,0 +1,193 @@
+package bloxid
+
+import (
+	"encoding/base32"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+const (
+	VersionUnknown Version = iota
+	Version0       Version = iota
+)
+
+type Version uint8
+
+func (v Version) String() string {
+	name, ok := version_name[v]
+	if !ok {
+		return version_name[VersionUnknown]
+	}
+	return name
+}
+
+var version_name = map[Version]string{
+	VersionUnknown: "unknown",
+	Version0:       "blox0",
+}
+
+var name_version = map[string]Version{
+	"unknown": VersionUnknown,
+	"blox0":   Version0,
+}
+
+var (
+	ErrInvalidVersion     error = errors.New("invalid bloxid version")
+	ErrInvalidEntityType  error = errors.New("entity type must be non-empty")
+	ErrInvalidUniqueIDLen error = errors.New("unique ID did not meet minimum length requirements")
+
+	ErrIDEmpty error = errors.New("empty bloxid")
+	ErrV0Parts error = errors.New("invalid number of parts found")
+)
+
+// NewV0 parse a string into a typed guid, return an error
+// if the string fails validation.
+func NewV0(bloxid string) (*V0, error) {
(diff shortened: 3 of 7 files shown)


bloxid/v0.go
```diff
@@ -0,0 +1,131 @@
...
+
+func randDefault() []byte {
+	return randBytes(DefaultEntropySize)
+}
+
+func randBytes(size int) []byte {
+	b := make([]byte, size/2)
+	_, err := rand.Read(b)
+	if err != nil {
+		panic(fmt.Sprintf("unable to read /dev/urandom: %s", err))
```
Evgeniy-L: It is better to have a more generic error here like `fail to generate random value: %s`. `unable to read /dev/urandom` may not be valid for some platforms.

drewwells: The error out of std library will clarify error, the method generally returns results from /dev/urandom but here's the verbose version: https://github.com/golang/go/blob/master/src/crypto/rand/rand.go#L14-L18

getrandom(2) http://man7.org/linux/man-pages/man2/getrandom.2.html

drewwells:
we have a flakey test, it's extra tough from the numerous versions of go being supported
```
ok  	github.com/infobloxopen/atlas-app-toolkit/rpc/resource	0.005s	coverage: 78.7% of statements

--- FAIL: TestServe (0.00s)

    --- FAIL: TestServe/closing_grpc_server_closes_http_server (0.00s)

        server_test.go:441: expected http server to be closed, but request was successfully sent

FAIL
```

drewwells:
fixed the failing test. I'm puzzled at why CI every passed, it failed 100% of the time locally.

bloxid/v0.go
Evgeniy-L: This code is in open source, so we should consider all platforms. For example if code is executed [On Windows systems, Reader uses the CryptGenRandom API](https://github.com/golang/go/blob/master/src/crypto/rand/rand.go#L17). In this case seeing `/dev/urandom` in error do not make any sense.  This is why we should avoid this and make message more generic. 
It is better even to just `panic(err)` here, because referring to `/dev/urandom` is not always valid.

bloxid/v0.go
```diff
@@ -0,0 +1,131 @@
...
+var (
+	ErrEmptyBloxID error = errors.New("empty bloxid")
+)
+
+func NewV0(bloxid string) (*V0, error) {
+	if len(bloxid) == 0 {
+		return nil, ErrEmptyBloxID
+	}
+	return nil, nil
+}
```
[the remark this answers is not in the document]
drewwells: it's complete now, with tests

bloxid/rand.go
```diff
@@ -0,0 +1,23 @@
...
+
+func randDefault() []byte {
+	return randBytes(DefaultEntropySize)
+}
+
+func randBytes(size int) []byte {
+	b := make([]byte, size/2)
+	_, err := rand.Read(b)
+	if err != nil {
+		panic(fmt.Sprintf("unable to read crypto store ie. /dev/urandom: %s", err))
```
drewwells:
```suggestion
		panic(err)
```

bloxid/interface.go
```diff
@@ -0,0 +1 @@
+package bloxid
```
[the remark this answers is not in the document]
drewwells: has contents now

Evgeniy-L:
Couple of questions to get more clarity:
1) Existing version of resource identifier had a special representation in proto - `atlas.rpc.Identifier`. How this resource identifier is going to be represented in proto?
2) Existing version of resource identifier is represented by special struct which has specific marshaling rules and DB codec which allows to have app/resource/{uuid} on rest and just {uuid} in db. Is something similar planned for new identifier?

@drewwells, as infobloxopen is an opensource repo we try not to post any links to internal docs here. I do not really like forcing `"blox0"` prefix in opensource. What do you think?

drewwells:
> Couple of questions to get more clarity:
> 
>     1. Existing version of resource identifier had a special representation in proto - `atlas.rpc.Identifier`. How this resource identifier is going to be represented in proto?
> 
>     2. Existing version of resource identifier is represented by special struct which has specific marshaling rules and DB codec which allows to have app/resource/{uuid} on rest and just {uuid} in db. Is something similar planned for new identifier?
> 
> 
> @drewwells, as infobloxopen is an opensource repo we try not to post any links to internal docs here. I do not really like forcing `"blox0"` prefix in opensource. What do you think?

Removed Google doc and copied some text from it. 

1. Let's do this in a separate PR as discussed
2. I think this makes sense based on Architecture discussion on 11/20

blox0 is a versioning scheme, so it makes sense to include it to me. Other users could create their own version separate from blox for their own usecases. It also may be useful for creating API clients. The API will generate a typed guid, and a client could parse it by using this package.
@daniel-garcia  thoughts open source?

Evgeniy-L:
@daniel-garcia, please share your thoughts about putting it to opensource

PR #181, opened 2019-12-05
Expose HandleRequestID

arashout:
Small PR to expose the `handleRequestID` function because it is useful.
I could just duplicate the code on my side, but I don't see a strong reason not to expose to function.


arashout:
I'm not really sure why the go1.12 version of the tests failed. It doesn't seem related to requestIDs

Issue #170, opened 2019-06-21
requestID middleware should entertain `X-Request-ID` along with `Request-ID` 


chinmayb:
Currently NGINX controller accepts `X-Request-ID` and auto generates an ID . Internal components who use requestID middleware looks for `Request-ID`. So the nginx requestID gets lost and new requestID gets propagated. 

Proposals
* make changes in nginx to accept `Request-ID` header
or
* Modify the middleware to accept just  ` X-Request-ID` or both.

kd7lxl:
It will be easier to modify the nginx config to accept `Request-ID`. I am working on a PR for this.

Evgeniy-L:
@chinmayb, @kd7lxl, this was investigated previously and there is a document on how to populate [Request Id At Nginx level](https://docs.google.com/document/d/1mP3AjZlbivh9ogli8ioThGkIll-lnFR4Bxh2Zc2cUmE/edit#heading=h.gf4dcsoy4w8y).

`request-id` middleware works on gRPC level and tries to extract Request-Id from gRPC metadata. See:
https://github.com/infobloxopen/atlas-app-toolkit/blob/master/requestid/requestid.go#L37
https://github.com/infobloxopen/atlas-app-toolkit/blob/master/gateway/header.go#L15

Headers that come from REST need to be translated to gRPC metadata. This is done by [DefaultHeaderMatcher](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L64). The problem here is that headers like `Request-ID` or `X-Request-ID` are not translated by default. Headers translation can solved like [this](https://github.com/infobloxopen/atlas-cli/pull/34/files).

Following the [Request Id At Nginx level](https://docs.google.com/document/d/1mP3AjZlbivh9ogli8ioThGkIll-lnFR4Bxh2Zc2cUmE/edit#heading=h.gf4dcsoy4w8y) doc we will be able to populate `Grpc-Metadata-Request-Id` header which will be translated to `Request-ID` into gRPC metadata by [DefaultHeaderMatcher](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L64). This way we will not need any changes for existing services. In a similar way we will be able to add other headers at Nginx level if needed.

drewwells:
[...]

PR #182, opened 2020-01-16
make requestid name consistent with JSON standart

burov:


chinmayb:
So are we not going to support `X-Request-Id` ?
https://github.com/infobloxopen/atlas-app-toolkit/issues/170

burov:
@chinmayb If you take a look into configuration for nginx we set it by Request-Id, and as well requested interceptor already doesn't support x-Request-Id

PR #183, opened 2020-02-03
Add an option to enable field mask override

burov:
* Implement option pattern for PresenceClientInterceptor
* Add WithOverrideFieldMask option to enable field mask override


gateway/field_presence.go
```diff
@@ -175,8 +192,10 @@ func PresenceClientInterceptor() grpc.UnaryClientInterceptor {
 
 		for i := 0; i < t.NumField(); i++ {
 			f := t.Field(i)
-			if f.Type() == reflect.TypeOf(fieldMask) && f.IsNil() {
-				f.Set(reflect.ValueOf(fieldMask))
+			if f.Type() == reflect.TypeOf(fieldMask) {
```
akleinib: Cannot you "return" after finding the field mask field? I assume there can be a single one in the request.

PR #188, opened 2020-04-21
Add common logger with specific fields

addudko:


logging/interceptor.go
```diff
@@ -39,6 +53,109 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
...
+	ctx = metadata.AppendToOutgoingContext(ctx, requestid.DefaultRequestIDKey, reqID)
+
+	if accountID, err := auth.GetAccountID(ctx, nil); err == nil {
+		fields[auth.MultiTenancyField] = accountID
+		ctx = metadata.AppendToOutgoingContext(ctx, auth.MultiTenancyField, accountID)
+	} else {
+		logger.Errorf("Unable to get %s from context", auth.MultiTenancyField)
+	}
+
+	if o.ophIDEnabled {
```
drewwells, line 129: this entire block is implementation specific and does not belong in the opensource project. I think you want to implement ways to instruct the toolkit to parse out specific fields, namely ophid. I believe this field will become less useful in the future as account ID is included ( in all cases ) for a JWT

logging/interceptor.go
```diff
@@ -39,6 +53,109 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
...
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		startTime := time.Now()
+		fields := newLoggerFields(method, startTime, clientKind)
+		err := invoker(ctx, method, req, reply, cc, opts...)
+		addSpecificFields(ctx, fields, logger, o)
+		code := o.codeFunc(err)
+		level := o.levelFunc(code)
+		durField, durVal := o.durationFunc(time.Since(startTime))
+		fields["grpc.code"] = code.String()
+		fields[durField] = durVal
```
burov, line 67: This block has poor readability , it will be better to reduce LOC between initialization and actual usage of variable, as well please separe each field by blank line 

```go
startTime := time.Now()

fields := newLoggerFields(method, startTime, clientKind)
addSpecificFields(ctx, fields, logger, o)

err := invoker(ctx, method, req, reply, cc, opts...)
if err != nil {
 	fields[logrus.ErrorKey] = err
}

code := o.codeFunc(err)
fields[GRPCCodeKey] = code.String()

durField, durVal := o.durationFunc(time.Since(startTime))
fields[durField] = durVal

level := o.levelFunc(code)
levelLogf(logrus.NewEntry(logger).WithFields(fields), level, "finished unary call with code "+code.String())

return err
```

logging/log.go
```diff
@@ -0,0 +1,82 @@
...
+	optCopy.levelFunc = grpc_logrus.DefaultCodeToLevel
+	for _, o := range opts {
+		o(optCopy)
+	}
+	return optCopy
+}
+
+func evaluateClientOpt(opts []Option) *options {
+	optCopy := &options{}
+	*optCopy = *defaultOptions
```
burov, line 76: From my point of view we can initialize defaultOptions per request instad of using global variable for this, this helps us to avoid such magic, simple is better than clever

logging/log.go
```diff
@@ -0,0 +1,82 @@
...
+		durationFunc: grpc_logrus.DefaultDurationToField,
+		ophIDEnabled: false,
+	}
+)
+
+type options struct {
+	levelFunc    grpc_logrus.CodeToLevel
+	shouldLog    grpc_logging.Decider
+	codeFunc     grpc_logging.ErrorToCode
+	durationFunc grpc_logrus.DurationToField
```
burov, line 53: I don't see this options can be changed by client, in this case i don't think this is a good idea to add them into options pattern

logging/log.go
```diff
@@ -0,0 +1,82 @@
...
+	ophIDEnabled bool
+}
+
+type Option func(*options)
+
+// Enables ophid field in interceptors logs
+func EnableOphID(opts *options) {
+	opts.ophIDEnabled = true
+}
```
burov, line 62: As i mentioned before, this is special case for a few our applications, suggest to avoid introducing this into common logger

logging/log.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"time"
+
+	grpc_logging "github.com/grpc-ecosystem/go-grpc-middleware/logging"
+	grpc_logrus "github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus"
+	"github.com/sirupsen/logrus"
+)
+
+func NewLogger(level string) *logrus.Logger {
```
burov, line 11: Should we instead of using simple string define custom type 
```go
type logLevel logrus.Level
```

and define a few public constansts 
```go
const (
   DebugLevel = logrus.DebugLevel
   InfoLevel = logrus.InfoLevel
   ...
)
```

This helps us to make this function safe and forbide clients pass bad level

logging/interceptor.go
addudko: Had the same thoughts regarding the ophid, but according to docs I understood this like we need logger with specific fields, set by default and not customizable.

logging/interceptor.go
```diff
@@ -39,6 +53,109 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
...
+		"grpc.service":          service,
+		"grpc.method":           method,
+		"grpc.start_time":       start.Format(time.RFC3339Nano),
+	}
+}
+
+func newLoggerForCall(ctx context.Context, entry *logrus.Entry, fields logrus.Fields) context.Context {
+	callLog := entry.WithFields(fields)
+	callLog = callLog.WithFields(ctxlogrus.Extract(ctx).Data)
+	return ctxlogrus.ToContext(ctx, callLog)
```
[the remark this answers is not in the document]
addudko: I'm not totally agree with this particular case, because it's perfectly readable I think. But I've changed it

logging/log.go
addudko: I think it's better to leave this argument as is and use already written levels from logrus. Don't see reasons to rewrite them

logging/interceptor.go
```diff
@@ -39,6 +60,133 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
...
+	return metadata.AppendToOutgoingContext(ctx, auth.MultiTenancyField, accountID), err
+}
+
+func addCustomField(ctx context.Context, fields logrus.Fields, logger *logrus.Logger, customField string) (context.Context, error) {
+	field, err := auth.GetJWTField(ctx, customField, nil)
+	if err != nil {
+		return ctx, fmt.Errorf("Unable to get custom %q field from context", customField)
+	}
+
+	fields[customField] = field
```
burov, line 164: Can we add the same option to get CustomFields from headers as well, from my point of view it can be reasonable

logging/gateway_interceptor.go
```diff
@@ -160,7 +162,7 @@ func GatewayLoggingInterceptor(logger *logrus.Logger, opts ...GWLogOption) grpc.
 		durField, durVal := grpc_logrus.DurationToTimeMillisField(time.Now().Sub(startTime))
 		fields = logrus.Fields{
 			durField:    durVal,
-			"grpc.code": grpc.Code(err).String(),
+			"grpc.code": status.Code(err).String(),
```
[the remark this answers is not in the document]
addudko: That not my implementation, should be deprecated if not used. Just updated deprecated field

logging/gateway_interceptor.go
```diff
@@ -169,7 +171,7 @@ func GatewayLoggingInterceptor(logger *logrus.Logger, opts ...GWLogOption) grpc.
 
 		// print log message with all fields
 		resLogger = resLogger.WithFields(fields)
-		levelLogf(resLogger, cfg.codeToLevel(grpc.Code(err)), "finished client unary call with code "+grpc.Code(err).String())
+		levelLogf(resLogger, cfg.codeToLevel(status.Code(err)), "finished client unary call with code "+status.Code(err).String())
```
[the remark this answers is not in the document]
addudko: Not my implementation, should be deprecated if not used. Just updated deprecated field

logging/interceptor.go
```diff
@@ -39,6 +60,133 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
...
+	ctx = metadata.AppendToOutgoingContext(ctx, customField, field)
+
+	return metadata.AppendToOutgoingContext(ctx, customField, field), err
```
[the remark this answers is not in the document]
addudko: My mistake, forget to remove

logging/interceptor.go
```diff
@@ -39,6 +60,133 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
...
+	}
+
+	return ctx
+}
+
+func addRequestIDField(ctx context.Context, fields logrus.Fields, logger *logrus.Logger, o *options) context.Context {
+	reqID, exists := requestid.FromContext(ctx)
+	if !exists || reqID == "" {
+		reqID = uuid.New().String()
+	}
```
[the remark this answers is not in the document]
Evgeniy-L: I agree with @burov, we should not spread the logic between multiple parties.
@addudko, please address this comment (e.g. return error similar to account_id).

logging/log.go
```diff
@@ -0,0 +1,92 @@
...
+	"google.golang.org/grpc/codes"
+)
+
+func New(level string) *logrus.Logger {
+	log := logrus.New()
+
+	// Default timeFormat
+	log.SetFormatter(&logrus.JSONFormatter{
+		TimestampFormat: time.RFC3339Nano,
+	})
```
kd7lxl, line 17: It would be helpful if there was a way to pass the logrus global logger here instead of always creating a new logger. This way we could apply these formatter and level settings globally for the process, which will take effect on all packages that use logrus, regardless of whether they accept a `logger` to their functions.

logging/interceptor.go
```diff
@@ -39,6 +62,161 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
...
+		levelLogf(
+			logrus.NewEntry(logger).WithFields(fields),
+			options.codeToLevel(code),
+			"finished unary call with code "+code.String())
+
+		return err
+	}
+}
+
+func UnaryServerInterceptor(logger *logrus.Logger, opts ...Option) grpc.UnaryServerInterceptor {
```
kd7lxl, line 91: If this were implemented as a [`ChainUnaryServer`](https://godoc.org/github.com/grpc-ecosystem/go-grpc-middleware#ChainUnaryServer), it could be composed of existing interceptors, like the requestid middleware, without having to reimplement them.

PR #189, opened 2020-04-29
Begin transaction with options.

maksim-volkov:
* Extended `Transaction` object with support of beginning transaction with
    options (to be able to set isolation level and RO option). Added
    `gorm.BeginTx()` wrapper;
  * Added analog wrapper to do same thing while extracting txn from context;
  * Updated UTs.

*Notes:*
1. `BeginWithOptions` used as name for new function while it wraps `gorm.BeginTx` maybe using `BeginTx` is preferable;
2. Aforementioned function only accept options and uses background context, based on [doc](https://golang.org/pkg/database/sql/#DB.BeginTx) this context (when cancelled) may be used to rollback transaction and prevent commit. I'm not sure if it should be exposed;  
3. `gorm/README.md` was not updated please let me know if it is required to change anything in it.

Verified with `make test`.

### DIFF
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -2,6 +2,7 @@ package gorm
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"reflect"
 	"sync"
@@ -71,6 +72,21 @@ func BeginFromContext(ctx context.Context) (*gorm.DB, error) {
 	return db, nil
 }
 
+func BeginWithOptionsFromContext(ctx context.Context, opts *sql.TxOptions) (*gorm.DB, error) {
+	txn, ok := FromContext(ctx)
+	if !ok {
+		return nil, ErrCtxTxnMissing
+	}
+	if txn.parent == nil {
+		return nil, ErrCtxTxnNoDB
+	}
+	db := txn.BeginWithOptions(opts)
+	if db.Error != nil {
+		return nil, db.Error
+	}
+	return db, nil
+}
+
 // Begin starts new transaction by calling `*gorm.DB.Begin()`
 // Returns new instance of `*gorm.DB` (error can be checked by `*gorm.DB.Error`)
 func (t *Transaction) Begin() *gorm.DB {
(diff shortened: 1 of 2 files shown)


gorm/transaction_test.go
```diff
@@ -120,32 +121,61 @@ func TestUnaryServerInterceptor_details(t *testing.T) {
...
+			mock.ExpectBegin()
+
+			gdb, err := gorm.Open("postgres", db)
+			if err != nil {
+				t.Fatalf("failed to open gorm db - %s", err)
+			}
+			txn := &Transaction{parent: gdb}
+
+			// test singleton behavior
+			opt := &sql.TxOptions{Isolation: sql.LevelSerializable}
```
Evgeniy-L, line 153: This can be moved under `true` case below as it is needed only it that case.

gorm/transaction_test.go
```diff
@@ -271,80 +301,148 @@ func TestContext(t *testing.T) {
...
+			ctxtxn := &Transaction{parent: gdb}
+			ctx = NewContext(ctx, ctxtxn)
+
+			var txn1, txn2 *gorm.DB
+			switch test.withOpts {
+			case true:
+				txn1, err = BeginWithOptionsFromContext(ctx, opt)
+			case false:
+				txn1, err = BeginFromContext(ctx)
+			}
```
Evgeniy-L, line 343: I do not really like this `switch` statements happening multiple times. Is it possible to pass function that creates transaction to t.Run instead of having multiple switch blocks?

gorm/transaction.go
```diff
@@ -71,6 +72,21 @@ func BeginFromContext(ctx context.Context) (*gorm.DB, error) {
 	return db, nil
 }
 
+func BeginWithOptionsFromContext(ctx context.Context, opts *sql.TxOptions) (*gorm.DB, error) {
```
Evgeniy-L, line 75: Can you please add some comments to this function

PR #190, opened 2020-05-12
Add stream interceptors

addudko:


logging/interceptor.go
```diff
@@ -80,23 +80,49 @@ func UnaryClientInterceptor(logger *logrus.Logger, opts ...Option) grpc.UnaryCli
 		fields[DefaultGRPCCodeKey] = code.String()
 
 		levelLogf(
-			logrus.NewEntry(logger).WithFields(fields),
+			entry.WithFields(fields),
 			options.codeToLevel(code),
-			"finished unary call with code "+code.String())
+			fmt.Sprintf("finished unary call with code %s", code.String()))
```
burov, line 85: Per definition of levelLogf ```func levelLogf(entry *logrus.Entry, level logrus.Level, format string, args ...interface{}) {``` we can add arguments directly into this function instead of using fmt.Sprintf

logging/interceptor.go
```diff
@@ -80,23 +80,49 @@ func UnaryClientInterceptor(logger *logrus.Logger, opts ...Option) grpc.UnaryCli
...
+			fields[logrus.ErrorKey] = err
+		}
+
+		code := status.Code(err)
+		fields[DefaultGRPCCodeKey] = code.String()
+
+		levelLogf(
+			entry.WithFields(fields),
+			options.codeToLevel(code),
+			fmt.Sprintf("finished client streaming call with code %s", code.String()))
```
burov, line 111: Per definition of levelLogf ```func levelLogf(entry *logrus.Entry, level logrus.Level, format string, args ...interface{}) {``` we can add arguments directly into this function instead of using fmt.Sprintf

logging/interceptor.go
```diff
@@ -109,12 +135,42 @@ func UnaryServerInterceptor(logger *logrus.Logger, opts ...Option) grpc.UnarySer
 		levelLogf(
 			ctxlogrus.Extract(newCtx).WithFields(fields),
 			options.codeToLevel(code),
-			"finished unary call with code "+code.String())
+			fmt.Sprintf("finished unary call with code %s", code.String()))
```
burov, line 138: Per definition of levelLogf ```func levelLogf(entry *logrus.Entry, level logrus.Level, format string, args ...interface{}) {``` we can add arguments directly into this function instead of using fmt.Sprintf

logging/interceptor.go
```diff
@@ -109,12 +135,42 @@ func UnaryServerInterceptor(logger *logrus.Logger, opts ...Option) grpc.UnarySer
...
+			fields[logrus.ErrorKey] = err
+		}
+
+		code := status.Code(err)
+		fields[DefaultGRPCCodeKey] = code.String()
+
+		levelLogf(
+			ctxlogrus.Extract(newCtx).WithFields(fields),
+			options.codeToLevel(code),
+			fmt.Sprintf("finished server streaming call with code %s", code.String()))
```
burov, line 168: Per definition of levelLogf ```func levelLogf(entry *logrus.Entry, level logrus.Level, format string, args ...interface{}) {``` we can add arguments directly into this function instead of using fmt.Sprintf

logging/interceptor.go
```diff
@@ -194,7 +276,7 @@ func addHeaderField(ctx context.Context, fields logrus.Fields, header string) (c
 
 	fields[strings.ToLower(header)] = field
 
-	return metadata.AppendToOutgoingContext(ctx, header, field), nil
+	return ctx, nil
```
burov, line 279: If we are not changing anything in context, can we remove it from return statement ?

logging/interceptor.go
```diff
@@ -181,9 +237,35 @@ func addCustomField(ctx context.Context, fields logrus.Fields, customField strin
...
+
+		fields[customField] = m
+
+		return ctx, err
+	}
+
 	fields[customField] = field
 
-	return metadata.AppendToOutgoingContext(ctx, customField, field), err
+	return ctx, err
```
burov, line 268: If we are not changing anything in context, can we remove it from return statement ?

logging/interceptor.go
```diff
@@ -172,7 +228,7 @@ func addAccountIDField(ctx context.Context, fields logrus.Fields) (context.Conte
 
 	fields[DefaultAccountIDKey] = accountID
 
-	return metadata.AppendToOutgoingContext(ctx, DefaultAccountIDKey, accountID), err
+	return ctx, err
```
burov, line 231: If we are not changing anything in context, can we remove it from return statement ?

logging/interceptor.go
```diff
@@ -161,7 +217,7 @@ func addRequestIDField(ctx context.Context, fields logrus.Fields) (context.Conte
 
 	fields[DefaultRequestIDKey] = reqID
 
-	return metadata.AppendToOutgoingContext(ctx, DefaultRequestIDKey, reqID), nil
+	return ctx, nil
```
burov, line 220: If we are not changing anything in context, can we remove it from return statement ?

logging/interceptor.go
addudko: That's a good point

logging/interceptor.go
```diff
@@ -62,14 +62,14 @@ func LogLevelInterceptor(defaultLevel logrus.Level) grpc.UnaryServerInterceptor
 	}
 }
 
-func UnaryClientInterceptor(logger *logrus.Logger, opts ...Option) grpc.UnaryClientInterceptor {
+func UnaryClientInterceptor(entry *logrus.Entry, opts ...Option) grpc.UnaryClientInterceptor {
 	options := initOptions(opts)
 
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		startTime := time.Now()
 		fields := newLoggerFields(method, startTime, DefaultClientKindValue)
```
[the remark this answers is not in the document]
addudko: No, if we're going to use client interceptor only for logging. 
Also try to imagine situation: we populate fields in client interceptor, put logger with fields into context, make rpc call and on the server side we may only extract logger from context and not making same job again above those fields which already present. So I think we shouldn't impose folow that way to server.

logging/mocks/server_stream.go
```diff
@@ -0,0 +1,433 @@
+// Code generated by counterfeiter. DO NOT EDIT.
+package mocks
```
Evgeniy-L, line 2: please add some details on how this mock was created, maybe put a readme

PR #193, opened 2020-06-13
Implement tracing package for toolkit

burov:
- [x] Refactoring for SpanContextAnnotator

- [x] Implement wrapper to extend default http plugin

- [x] Implement wrapper to extend default gRPC plugin

- [x] Util functions CreateSpan, GetCurrent, AddLogs, AddAttributes
- [x] Builder for exporter 

- [x] Add grpc error into span if exists

- [x] Obfuscation for headers with sensitive data 

- [ ] Automated tests

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -40,7 +40,7 @@
 
 [[constraint]]
    name = "go.opencensus.io"
-   version = "0.22.3"
+   version = ">=0.22.3"
 
 [prune]
   go-tests = true
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -38,6 +38,10 @@
   name = "google.golang.org/grpc"
   version = ">=1.0.0"
 
+[[constraint]]
+   name = "go.opencensus.io"
+   version = "0.22.3"
+
 [prune]
   go-tests = true
   unused-packages = true
--- a/tracing/grpc.go
+++ b/tracing/grpc.go
@@ -1,0 +1,188 @@
+package tracing
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus/ctxlogrus"
+	"go.opencensus.io/plugin/ocgrpc"
+	"go.opencensus.io/trace"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/stats"
+)
+
+//GRPCOption allows extending handler with additional functionality
+type GRPCOption func(*gRPCOptions)
+
+type metadataMatcher func(string) (string, bool)
+
+type gRPCOptions struct {
+	spanWithMetadata func(context.Context, stats.RPCStats) bool
+	metadataMatcher  metadataMatcher
+
+	spanWithPayload func(context.Context, stats.RPCStats) bool
+	maxPayloadSize  int
+}
+
+func defaultGRPCOptions() *gRPCOptions {
+	return &gRPCOptions{
+		metadataMatcher: defaultMetadataMatcher,
+		maxPayloadSize:  DefaultMaxPayloadSize,
+
+		//Keep spanWithMetadata and spanWithPayload equals to nil instead of dummy functions
+		//to prevent path trough header for each request
+	}
+}
+
+//WithMetadataAnnotation annotate span with request metadata
+func WithMetadataAnnotation(f func(context.Context, stats.RPCStats) bool) GRPCOption {
+	return func(ops *gRPCOptions) {
+		ops.spanWithMetadata = f
+	}
+}
+
+//WithMetadataMatcher set metadata matcher to filterout or preprocess metadata
+func WithMetadataMatcher(f func(string) (string, bool)) GRPCOption {
+	return func(ops *gRPCOptions) {
+		ops.metadataMatcher = f
+	}
+}
+
+//WithGRPCPayloadAnnotation add Inbound/Outbound payload as an attribute to span if f returns true
+func WithGRPCPayloadAnnotation(f func(context.Context, stats.RPCStats) bool) GRPCOption {
+	return func(ops *gRPCOptions) {
+		ops.spanWithPayload = f
+	}
+}
+
+//WithGRPCPayloadLimit limit payload size propogated to span
+//in case payload exceeds limit, payload truncated and
+//annotation payload.truncated=true added into span
+func WithGRPCPayloadLimit(limit int) GRPCOption {
+	return func(ops *gRPCOptions) {
+		ops.maxPayloadSize = limit
+	}
+}
+
+//Check that &ServerHandler{} comply stats.Handler interface
+var _ stats.Handler = &ServerHandler{}
+
+//ServerHandler is a wrapper over ocgrpc.ServerHandler
+//wrapper extends metadata added into the  span
+type ServerHandler struct {
+	ocgrpc.ServerHandler
+
+	options *gRPCOptions
+}
+
+//NewServerHandler returns wrapper over ocgrpc.ServerHandler
+func NewServerHandler(ops ...GRPCOption) *ServerHandler {
+	options := defaultGRPCOptions()
+	for _, op := range ops {
+		op(options)
+	}
+
+	return &ServerHandler{options: options}
+}
+
+// HandleRPC implements per-RPC tracing and stats instrumentation.
+func (s *ServerHandler) HandleRPC(ctx context.Context, rs stats.RPCStats) {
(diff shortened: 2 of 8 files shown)


tracing/http.go
```diff
@@ -0,0 +1,228 @@
...
+	ResponsePayloadAnnotationKey = "response.payload"
+
+	//DefaultMaxPayloadSize represent max payload size which will be added to span
+	DefaultMaxPayloadSize = 1024 * 1024
+
+	//ReducedMarkerKey is a key for annotation which will be presented in span in case payload was reduced
+	ReducedMarkerKey = "reduced_payload"
+
+	//ReducedMarkerValue is a value for annotation which will be presented in span in case payload was reduced
+	ReducedMarkerValue = "true"
```
Evgeniy-L, line 34: is it a good idea to use "true" string here? It is something we invented or values is forced by 3rd party?

tracing/http.go
```diff
@@ -0,0 +1,228 @@
...
+	RequestPayloadAnnotationKey = "request.payload"
+
+	//ResponsePayloadAnnotationKey is a key under which response payload stored in span
+	ResponsePayloadAnnotationKey = "response.payload"
+
+	//DefaultMaxPayloadSize represent max payload size which will be added to span
+	DefaultMaxPayloadSize = 1024 * 1024
+
+	//ReducedMarkerKey is a key for annotation which will be presented in span in case payload was reduced
+	ReducedMarkerKey = "reduced_payload"
```
Evgeniy-L, line 31: does it make same to have this key consistent to keys above ("." vs "_"). Maybe use `payload.reduced`.
It is something we invented or key is forced by 3rd party?

tracing/http.go
```diff
@@ -0,0 +1,228 @@
...
+
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/trace"
+)
+
+const (
+	//RequestHeaderAnnotationPrefix is a prefix which added to each request header attribute
+	RequestHeaderAnnotationPrefix = "request.header."
+
+	//ResponseHeaderAnnotationPrefix is a prefix which added to each response header attribute
```
Evgeniy-L, line 18:
```suggestion
	//ResponseHeaderAnnotationPrefix is a prefix which is added to each response header attribute
```

tracing/http.go
```diff
@@ -0,0 +1,228 @@
...
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/trace"
+)
+
+const (
+	//RequestHeaderAnnotationPrefix is a prefix which added to each request header attribute
```
Evgeniy-L, line 15:
```suggestion
	//RequestHeaderAnnotationPrefix is a prefix which is added to each request header attribute
```

tracing/http.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (w *responseBodyWrapper) Write(b []byte) (int, error) {
+	//In case we recieve an error from Writing into buffer we just skip it
+	//because adding payload to span is not so critical as provide response
+	_, _ = w.buffer.Write(b)
+	return w.ResponseWriter.Write(b)
+}
+
+func markSpanReduced(s *trace.Span) {
+	s.AddAttributes(trace.StringAttribute("reduced_payload", "true"))
```
Evgeniy-L, line 205:
```suggestion
	s.AddAttributes(trace.StringAttribute(ReducedMarkerKey, ReducedMarkerValue))
```

Gopkg.toml
```diff
@@ -38,6 +38,10 @@
   name = "google.golang.org/grpc"
   version = ">=1.0.0"
 
+[[constraint]]
+   name = "go.opencensus.io"
+   version = "0.22.3"
```
Evgeniy-L, line 43: please use >= for version

tracing/http.go
burov: It our own marker which we add to span with reduced body, to allow developers to filter out them, if they want to see spans only with full body, for me "true" good, any idea how to do it more descriptive ?

tracing/http.go
burov: Thank you, that a good idea

Gopkg.toml
burov: Sure

tracing/http.go
Evgeniy-L: I was thinking about using "reduced" or "enabled" string instead of "true". I am fine with any option (including existing), so it is up to you.

burov: Actually we can discuss it, is it usefull to add some marker that span contains payload and signal not only that payload reduced but signal that payload actually exists for example payload.type = reduced|enabled allow developers to have more granural filtering, we will think about it

tracing/http.go
```diff
@@ -0,0 +1,228 @@
...
+
+//WithPayloadSize ...
+func WithPayloadSize(maxSize int) Option {
+	return func(ops *options) {
+		ops.maxPayloadSize = maxSize
+	}
+}
+
+//NewTracingMiddleware wrap handler
+func NewTracingMiddleware(ops ...Option) func(http.Handler) http.Handler {
```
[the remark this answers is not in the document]
burov: Good point

burov:
> Are the UnaryServerInterceptor and StreamingServerInterceptor next?

Actually no, we decided to leverage already existed plugins from open census and from our side just wrap them, and inreach them with additional functionality which important for us.

tracing/http.go
mphillips-infoblox: Are we compressing the payload ("reduced"), truncating (simply cut off after X characters and inject an ellipse to visually indicate truncation), or replacing the data (replacing data with a marker indicating that data has been replaced)

burov: That a good point @mphillips-infoblox, i will change reduced to truncated actually, I already put "..." in the body but label for me seems reasonable to add annotation as well, in case somebody for example what to filterout traces without full body

mphillips-infoblox: Yes, as @Evgeniy-L suggested, if you are consistent and use something like payload.truncated = true, that would work really well

tracing/http.go
```diff
@@ -0,0 +1,239 @@
...
+var _ http.Handler = &Handler{}
+
+//Handler is a opencensus http plugin wrapper which do some usefull things to reach traces
+type Handler struct {
+	child http.Handler
+
+	options *httpOptions
+}
+
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
drewwells, line 125: What use case is this covering? If I make a grpc call to the service, it won't run this code. Wouldn't it be simpler to handle the gRPC case and not bother implementing the http Handler. We only need gateway to pass along headers to the underlying gRPC server

tracing/grpc.go
```diff
@@ -0,0 +1,188 @@
...
+
+//NewServerHandler returns wrapper over ocgrpc.ServerHandler
+func NewServerHandler(ops ...GRPCOption) *ServerHandler {
+	options := defaultGRPCOptions()
+	for _, op := range ops {
+		op(options)
+	}
+
+	return &ServerHandler{options: options}
+}
```
drewwells, line 89: How do I read the span from the context.Context passed into middleware? For example, authZ reads a span from the passed context to its middleware then creates additional spans inside of it to trace OPA. See here: https://github.com/Infoblox-CTO/ngp.authz/pull/745/files#diff-fd5b75f59732938618e94d5a57a06c35R112

tracing/http.go
mphillips-infoblox: You are correct, however we want to be able to trace the call flow at all the protocol boundaries for a service.  This allows us to not only see what the call properties are (headers, payload, etc) but we can also collect timing associated with the call pipeline for any entry point into/out of the service.

tracing/grpc.go
mphillips-infoblox: There are helper methods in this toolkit that take care of the plumbing for you.  You just request a span passing in the context.

tracing/grpc.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, false, err
+	}
+
+	bytes, truncated := truncatePayload(bytes, limit)
+	attrs := []trace.Attribute{trace.StringAttribute(key, string(bytes))}
+	return attrs, truncated, nil
+}
+
+func obfuscate(x string) string {
+	countChars := int(float64(len(x)) * 0.20)
```
mphillips-infoblox, line 197: This should probably be a constant.  We should consider making this configurable in the future.

mphillips-infoblox: The 0.20 value

PR #196, opened 2020-07-22
B-27963: Util to Include Headers into outgoing context

sstaratsitarau:


util/context.go
```diff
@@ -0,0 +1,43 @@
...
+	if rid, ok := requestid.FromContext(ctx); ok {
+		params[requestid.DefaultRequestIDKey] = []string{rid}
+	}
+
+	if rIPs, ok := gateway.Header(ctx, gateway.XForwardedFor); ok {
+		params[gateway.XForwardedFor] = []string{rIPs}
+	}
+
+	return ForwardContextWithCustomParams(ctx, params)
+}
```
kd7lxl, line 29: Do tracing span-ids need to be forwarded here?

util/context.go
```diff
@@ -0,0 +1,43 @@
...
+
+// ForwardContextWithCustomParams set to outgoing context all params from incoming context + custom params
+func ForwardContextWithCustomParams(ctx context.Context, params map[string][]string) context.Context {
```
kd7lxl, line 32: I'm not sure this function is needed. Usage of [google.golang.org/grpc/metadata](https://godoc.org/google.golang.org/grpc/metadata) is nearly as simple:
```go
md, _ := metadata.FromIncomingContext(ctx)
md = md.Join(params)
outgoingCtx := metadata.NewOutgoingContext(context.Background(), md)
```

gateway/header.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 )
 
+const (
+	// XForwardedFor is a header to identify the origin of an IP address
+	XForwardedFor = "X-Forwarded-For"
+)
```
burov, line 18: What the reason to keep this constant here ?

util/context.go
```diff
@@ -0,0 +1,54 @@
...
+	if !ok {
+		return ctx
+	}
+
+	for key := range params {
+		md[key] = params[key]
+	}
+
+	return metadata.NewOutgoingContext(ctx, md)
+}
```
burov, line 54: I don't think we really need this function

util/context.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/infobloxopen/atlas-app-toolkit/requestid"
+)
+
+// ForwardContext set to outgoing context request_id, auth_token and X-Forwarded-For header value
+func ForwardContext(ctx context.Context) context.Context {
+	keys := []string{
+		auth.AuthorizationHeader,
+		requestid.DefaultRequestIDKey,
+		gateway.XForwardedFor,
+	}
```
burov, line 19: Probably add tracing context header will be good as well, see https://github.com/census-instrumentation/opencensus-go/blob/v0.22.4/trace/trace.go#L113 for more information

gateway/header.go
sstaratsitarau: This is the key for the "X-Forwarded-For" value, which is the http request header. I figured the gateway package and header.go file would the most suitable place for this variable.

Could you suggest a better location for this variable?

util/context.go
```diff
@@ -0,0 +1,54 @@
...
+
+	return ForwardContextWithCustomParams(ctx, keys)
+}
+
+// ForwardContextWithCustomParams set to outgoing context specified parameters from incoming context by keys
+func ForwardContextWithCustomParams(ctx context.Context, keys []string) context.Context {
```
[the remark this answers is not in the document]
sstaratsitarau: I don't quite understand why. We just pass keys, which are string values. For what reason should this way be changed to variadic parameters?

util/context.go
Evgeniy-L: let remove this function
we can consider adding it if requested later

addudko:
Closed due to 
https://github.com/infobloxopen/atlas-app-toolkit/pull/211

PR #197, opened 2020-07-23
Fix request id key

burov:
Fix request id key to conform with https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields

### DIFF
--- a/logging/interceptor_test.go
+++ b/logging/interceptor_test.go
@@ -228,6 +228,50 @@ func TestUnaryServerInterceptor(t *testing.T) {
 	assert.Equal(t, "finished unary call with code OK", result["msg"])
 }
 
+func TestUnaryServerInterceptorDeprecatedHeader(t *testing.T) {
+	testLogger.Out = &buf
+	interceptor := UnaryServerInterceptor(logrus.NewEntry(testLogger))
+
+	md := metautils.NiceMD{}.Set(testAuthorizationHeader, testJWT).Set(requestid.DeprecatedRequestIDKey, testRequestID).Set(testCustomHeaderKey, testCustomHeaderVal)
+	ctx := metadata.NewIncomingContext(context.Background(), metadata.MD(md))
+
+	handlerMock := func(ctx context.Context, req interface{}) (interface{}, error) {
+		newMD, ok := metadata.FromIncomingContext(ctx)
+		assert.True(t, ok)
+		assert.Equal(t, testJWT, newMD.Get(testAuthorizationHeader)[0])
+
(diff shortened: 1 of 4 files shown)


requestid/requestid.go
```diff
@@ -14,7 +14,8 @@ import (
 // DefaultRequestIDKey is the metadata key name for request ID
 const (
 	DeprecatedRequestIDKey = "Request-Id"
-	DefaultRequestIDKey    = "request_id"
+	DefaultRequestIDKey    = "X-Request-ID"
```
kd7lxl, line 17: Note that this is already implemented in the ingress-controller:
https://github.com/Infoblox-CTO/athena.devops/blob/d9556173c9d749f86515876cadeb6fa0ec9a60d8/cluster-provisioning/charts/base/files/nginx.tmpl#L322-L329

But not forwarded to identity:
https://github.com/Infoblox-CTO/athena.devops/blob/d9556173c9d749f86515876cadeb6fa0ec9a60d8/cluster-provisioning/charts/base/files/nginx.tmpl#L920
(on auth requests, it's set as `Request-ID` only)

logging/interceptor_test.go
```diff
@@ -38,7 +39,7 @@ var (
 	buf         bytes.Buffer
 	reader      io.Reader
 	testLogger  = New("Info")
-	testMD      = metautils.NiceMD{}.Set(testAuthorizationHeader, testJWT).Set(DefaultRequestIDKey, testRequestID).Set(testCustomHeaderKey, testCustomHeaderVal)
+	testMD      = metautils.NiceMD{}.Set(testAuthorizationHeader, testJWT).Set(requestid.DefaultRequestIDKey, testRequestID).Set(testCustomHeaderKey, testCustomHeaderVal)
```
kd7lxl, line 42: Please add a case that looks for `DeprecatedRequestIDKey` so we can verify backwards compatibility.

requestid/requestid.go
Evgeniy-L: @kd7lxl, please avoid references to data from private repos in public repositories

burov: @kd7lxl We probably can forward it, it a good point

logging/interceptor_test.go
burov: Good point, did it

requestid/requestid.go
drewwells: broke authz middleware as well https://github.com/Infoblox-CTO/ngp.authz.middleware/pull/23/files#diff-320a504f903c9c37cd1c4852412110f8R216

PR #198, opened 2020-08-16
convert to go mod

drewwells:
Many of our projects have switched to go mod. In order to import toolkit accurately, it needs to use go modules.


Evgeniy-L:
Migration to go mod is done with #208

PR #201, opened 2020-10-13
Provide customization to allow passing X-Geo-* headers into gRPC metadata

burov:


Evgeniy-L (approved):
Looks good

gateway/header.go
```diff
@@ -114,16 +114,53 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
...
 func ExtendedDefaultHeaderMatcher(headerNames ...string) func(string) (string, bool) {
 	customHeaders := map[string]bool{}
 	for _, name := range headerNames {
 		customHeaders[strings.ToLower(name)] = true
 	}
+
+	atlasMatcher := AtlasDefaultHeaderMatcher()
 	return func(headerName string) (string, bool) {
-		if key, ok := runtime.DefaultHeaderMatcher(headerName); ok {
+		if key, ok := atlasMatcher(headerName); ok {
```
kd7lxl, line 163: Rather than plugging the new headers in like this, I think it would be better to implement
```go
func ChainHeaderMatcher(matchers ...func(string) (string, bool))  func(string) (string, bool)
```
so that these custom matches can be used both independently and together, not just together. It's the same pattern we use to implement interceptors, instead of wrapping interceptors.

burov: It actually a good idea,i will add ChainHeaderMatcher

Evgeniy-L (approved):

gateway/header.go
kd7lxl: I provided an example of how `ChainHeaderMatcher()` can simplify this code in #202

PR #208, opened 2021-01-16
Migrate atlas toolkit to go mod

alavrovinfb:
migrated to go mod

updated travis CI


go.mod
```diff
@@ -0,0 +1,32 @@
...
+	github.com/stretchr/testify v1.5.1
+	go.opencensus.io v0.22.3
+	golang.org/x/net v0.0.0-20200520182314-0ba52f642ac2
+	google.golang.org/api v0.26.0 // indirect
+	google.golang.org/genproto v0.0.0-20200527145253-8367513e4ece
+	google.golang.org/grpc v1.29.1
+	gopkg.in/yaml.v2 v2.2.8 // indirect
+)
+
+replace (
```
alavrovinfb, line 27: The following packages are locked according to versions from Gopkg.lock. `Go mod tidy` tries to upgrade the mentioned packages to newer versions, although, these versions aren't major, looks like they contain breaking changes.
As result, TestForwardResponseMessageWithNil test fails with the following error message:
```
--- FAIL: TestForwardResponseMessageWithNil (0.00s)
panic: interface conversion: interface {} is nil, not map[string]interface {} [recovered]
	panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 62 [running]:
testing.tRunner.func1.1(0x936f20, 0xc000324450)
	/usr/local/go/src/testing/testing.go:1072 +0x30d
```
https://github.com/infobloxopen/atlas-app-toolkit/blob/55b3a20b5562a04274a9e6cb8e58a432e398a993/gateway/response_test.go#L271

So, if we want to upgrade to the latest version, we need to investigate the potential impact on our codebase.

alavrovinfb:
CI passed.

Evgeniy-L (approved):
LGTM

PR #210, opened 2021-01-20
Added support for graceful shutdown

prybintsev:
Added support for disabling automatic server stop
Added support for graceful shutdown

Due to the following line in "Serve" function:

`defer s.Stop()`

It was impossible to perform a graceful shutdown of the services, since execution of graceful shutdown functions s.GRPCServer.GracefulStop() and s.HTTPServer.Shutdown(ctx) was causing "Server" function to return and s.Stop to be called which terminates the servers ungracefully.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -39,6 +39,8 @@ type Server struct {
 
 	// HTTPServer will be started whenever this is served
 	HTTPServer *http.Server
+
+	isAutomaticStop bool
 }
 
 //Middleware wrapper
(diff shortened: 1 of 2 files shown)


prybintsev:
Hi @Evgeniy-L,

This is a PR with the changes we have discussed by email

prybintsev:
Hi @Evgeniy-L 

Will you get a chance to take a look at this PR?

Evgeniy-L (approved):

PR #211, opened 2021-01-20
Refactor Headers. Util to include Headers into outgoing context

addudko:


auth/jwt.go
```diff
@@ -13,6 +13,9 @@ const (
 	// MultiTenancyField the field name for a specific tenant
 	MultiTenancyField = "account_id"
 
+	// AuthorizationHeader contains information about the header value for the token
+	AuthorizationHeader = "Authorization"
```
drewwells, line 17: 👍🏻

gateway/header.go
```diff
@@ -12,6 +12,33 @@ import (
...
+		"x-geo-country-code",
+		"x-geo-country-name",
+		"x-geo-region-code",
+		"x-geo-region-name",
+		"x-geo-city-name",
+		"x-geo-postal-code",
+		"x-geo-latitude",
+		"x-geo-longitude",
+	}
+}
```
drewwells, line 30: Why is this a func? If it's meant to make it extendable, a variable would just as well

```suggestion
var GetGeoHeaders = []string{
		"x-geo-org",
		"x-geo-country-code",
		"x-geo-country-name",
		"x-geo-region-code",
		"x-geo-region-name",
		"x-geo-city-name",
		"x-geo-postal-code",
		"x-geo-latitude",
		"x-geo-longitude",
}
```

gateway/header.go
```diff
@@ -12,6 +12,33 @@ import (
...
+
+// GetXB3Headers returns a slice of x-b3- headers.
+func GetXB3Headers() []string {
+	return []string{
+		"x-b3-traceid",
+		"x-b3-parentspanid",
+		"x-b3-spanid",
+		"x-b3-sampled",
+	}
+}
```
drewwells, line 40: Same comment
```suggestion
var GetXB3Headers = []string{
		"x-b3-traceid",
		"x-b3-parentspanid",
		"x-b3-spanid",
		"x-b3-sampled",
	}
}
```

gateway/header.go
```diff
@@ -116,17 +143,7 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
...
-		"x-geo-country-code",
-		"x-geo-country-name",
-		"x-geo-region-code",
-		"x-geo-region-name",
-		"x-geo-city-name",
-		"x-geo-postal-code",
-		"x-geo-latitude",
-		"x-geo-longitude",
-	)
+	return ExtendedDefaultHeaderMatcher(GetGeoHeaders()...)
```
drewwells, line 146:
```suggestion
	return ExtendedDefaultHeaderMatcher(GetGeoHeaders...)
```

gateway/header.go
```diff
@@ -136,12 +153,7 @@ func RequestIDHeaderMatcher() runtime.HeaderMatcherFunc {
 
 // TracingHeaderMatcher tracing headers
 func TracingHeaderMatcher() runtime.HeaderMatcherFunc {
-	return ExtendedDefaultHeaderMatcher(
-		"x-b3-traceid",
-		"x-b3-parentspanid",
-		"x-b3-spanid",
-		"x-b3-sampled",
-	)
+	return ExtendedDefaultHeaderMatcher(GetXB3Headers()...)
```
drewwells, line 156:
```suggestion
	return ExtendedDefaultHeaderMatcher(GetXB3Headers...)
```

util/context.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+	"google.golang.org/grpc/metadata"
+
+	"github.com/infobloxopen/atlas-app-toolkit/auth"
+	"github.com/infobloxopen/atlas-app-toolkit/gateway"
+	"github.com/infobloxopen/atlas-app-toolkit/requestid"
+)
+
+// ForwardContext set to outgoing context request_id, auth_token and X-Forwarded-For header value
+func ForwardContext(ctx context.Context) context.Context {
```
drewwells, line 13: I think OutgoingContext would be a better name here. These are smaller helpers around metadata.NewOutgoingContext()

drewwells (approved):
Few comments, looks fine.

gateway/header.go
addudko: The reason why I made it as a func is assuming that this headers are meant to be constant. Such approach ensures that there is no way we can get mutated value.

util/context.go
addudko: Agree

Evgeniy-L (approved):

PR #229, opened 2021-03-02
Removes cyclic type references

drewwells:
grpc-gateway no longer allows cyclic references. Creates a cyclic
    reference between two types to avoid the check

    regenerated pb from latest atlas-gentool
    nested type

unblocks #1151


query/collection_operators.proto
```diff
@@ -71,24 +71,55 @@ message Filtering {
...
 // left and right are respectively left and right operands of the operator, could be
 // either LogicalOperator or one of the supported conditions.
 // is_negative is set to true if the operator is negated.
 message LogicalOperator {
     oneof left {
-        LogicalOperator left_operator = 1;
+        // cyclic type definitions are no longer available in grpc-gateway
+        // https://github.com/grpc-ecosystem/grpc-gateway/pull/1266
+        // TODO: add a nested LogicalOperator if deeper nesting of query parameters is necessary
+        NestedLogicalOperator left_operator = 1;
```
drewwells, line 114: Avoids the cyclic checks by flipping between two types

drewwells: This does not actually work. The test server doesn't reproduce the problem, but authz proto saw the same cyclic detection get alerted.

drewwells:
after discussion, we're going to rebase atlas-patch on top of 3years of upstream changes 😓

PR #243, opened 2021-04-15
Add options to transaction interceptor.

amikhailau:
We are trying to optimize one of our services readiness check by introducing checks based on existing transactions. For that we require introduction of options for transaction interceptor.

### DIFF
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -174,9 +174,13 @@ func (t *Transaction) Commit(ctx context.Context) error {
 // If call of grpc.UnaryHandler returns with an error the transaction
 // is aborted, otherwise committed.
 func UnaryServerInterceptor(db *gorm.DB) grpc.UnaryServerInterceptor {
+	txn := &Transaction{parent: db}
+	return UnaryServerInterceptorTxn(txn)
+}
+
+func UnaryServerInterceptorTxn(txn *Transaction) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
 		// prepare new *Transaction instance
-		txn := &Transaction{parent: db}
 
 		defer func() {
 			// simple panic handler


gorm/transaction.go
```diff
@@ -173,10 +173,15 @@ func (t *Transaction) Commit(ctx context.Context) error {
...
-func UnaryServerInterceptor(db *gorm.DB) grpc.UnaryServerInterceptor {
+func UnaryServerInterceptor(db *gorm.DB, opts ...func(context.Context, interface{}, *Transaction)) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
 		// prepare new *Transaction instance
 		txn := &Transaction{parent: db}
+		if len(opts) > 0 {
+			for _, opt := range opts {
+				opt(ctx, req, txn)
+			}
+		}
```
drewwells, line 184: Please describe your use case, tests, and/or examples. It is not clear from this Pr what you are trying to accomplish. 

The functional option is very limited in scope. It would be better to implement flexible options where this callback behavior is just one type. See example below

```
func WithCallback(func(context.Context, interface{}, *Transaction) error) Option {
}

func UnaryServerInterceptor(db,  opts ...Option) grpc.UnaryServerInterceptor {
...
}
```

drewwells (changes requested):
Please provide background for what you are trying to accomplish. This would introduce API changes that we can not mutate in the future easily.

gorm/transaction.go
wplu: use case PR messaged via DM.  Drew's function sig is much more flexible.

gorm/transaction.go
```diff
@@ -173,10 +175,18 @@ func (t *Transaction) Commit(ctx context.Context) error {
 // Client is responsible to call `txn.Begin()` to open transaction.
 // If call of grpc.UnaryHandler returns with an error the transaction
 // is aborted, otherwise committed.
-func UnaryServerInterceptor(db *gorm.DB) grpc.UnaryServerInterceptor {
+func UnaryServerInterceptor(db *gorm.DB, opts ...Option) grpc.UnaryServerInterceptor {
```
drewwells, line 178: Try adding a new func that takes a Transaction. You would then have the ability to register a hook. 
```
 txn := NewTransaction(db)
 txn.AddAfterCommitHook(...myhook...)
  
 UnaryServerInterceptorTxn(txn) grpc.UnaryServerInterceptor {
 ...
 }
```

drewwells (changes requested):
See interface comment

wplu (approved):

drewwells (approved):

PR #247, opened 2021-05-13
Remove default header matcher.

amikhailau:


Evgeniy-L (approved):

kd7lxl:
This looks like it will be a breaking change for many apps. Please share more details.

amikhailau:
> This looks like it will be a breaking change for many apps. Please share more details.

When we updated atlas-app-toolkit version in athena-identity we noticed that some headers are not reaching the server and located issue described here: https://infoblox.atlassian.net/browse/ATLAS-10018
Please take a note that this PR is not targeting master, but only a separate branch that we use in athena-identity

kd7lxl:
Okay, further investigation: https://github.com/infobloxopen/atlas-app-toolkit/commit/bb98be744ae09fd35c79bce56827b6a453e8c181#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4L138-R143 introduced a regression that replaced the header matcher in the provided options.

Best practice is to explicitly call the `DefaultHeaderMatcher()`: https://github.com/infobloxopen/atlas-cli/blob/4d2d5d2855c7dfdb242ec8c4e68369abe00bcf02/atlas/templates/cmd/server/main.go.gotmpl#L132 which is compatible with the change in this PR.

Since this PR is not to master branch, it looks like this regression still exists on master, but fixing it could be a breaking change for any app that did not follow the best practice example.

PR #251, opened 2021-05-26
Add validation for sort criteria

emilev-ms:


query/sorting.go
```diff
@@ -29,7 +30,10 @@ func (c SortCriteria) GoString() string {
 // See: https://github.com/infobloxopen/atlas-app-toolkit#sorting
 func ParseSorting(s string) (*Sorting, error) {
 	var sorting Sorting
-
+	sortingRegexp := regexp.MustCompile("^[A-Za-z0-9_,. ]+$")
```
Evgeniy-L, line 33: This change makes sense. I just would like to clarify why this set of allowed characters is choosen?
I think we should also allow `-` (maybe something else?)

EvgenyMilev: Updated the regexp pattern, added  dash.
As for other special characters, the following are allowed:
1. Any letters and numbers
2. Underscore as it can be used in the object name
3. Comma as a separator in the array of sorting criteria.
4. Dot that can be used for nested objects

Evgeniy-L (approved):

query/sorting.go
```diff
@@ -29,7 +30,10 @@ func (c SortCriteria) GoString() string {
 // See: https://github.com/infobloxopen/atlas-app-toolkit#sorting
 func ParseSorting(s string) (*Sorting, error) {
 	var sorting Sorting
-
+	sortingRegexp := regexp.MustCompile(`^[A-Za-z0-9_,.\- ]+$`)
+	if !sortingRegexp.MatchString(s) {
+		return nil, fmt.Errorf("invalid character in the sort criteria")
+	}
```
drewwells, line 36: I see no explanation here as to why this is being checked https://github.com/infobloxopen/atlas-app-toolkit/tree/master/query#sorting 
Why is this needed?

EvgenyMilev: The purpose of these changes is to exclude the use of special characters in the sorting criteria. Otherwise, a similar validation must be used in every List method where a * query.Sorting object is used. For example, such a request .../v2/users? _Limit5 & offset = 5 & order_by = <string> will return an error like "internal-error pq: syntax error at or near" <"". But I would like to have the error "Invalid request"

PR #265, opened 2021-07-13
Added fast fail option into checksHandler

nikonok:


health/handler.go
```diff
@@ -25,7 +28,7 @@ type Checker interface {
 
 // NewChecksHandler accepts two strings: health and ready paths.
 // These paths will be used for liveness and readiness checks.
-func NewChecksHandler(healthPath, readyPath string) Checker {
+func NewChecksHandler(healthPath, readyPath string, failFast bool) Checker {
```
Evgeniy-L, line 31: This seem to be a breaking change. E.g. every app after updating to new version will need to change
`handler := NewChecksHandler("/healthz", "/ready")` to `handler := NewChecksHandler("/healthz", "/ready", false)`.

We should avoid this. Options might be:
1.  separate new constructor
2.  function in handle to set fail fast method
3.  ability for pass options to constrictor as we have [here](https://github.com/infobloxopen/atlas-app-toolkit/blob/9665be2a3ad4fef84696a8a0fc7224c8c3f6f135/gateway/gateway.go#L97)
4.  whatever you think reasonable

Evgeniy-L (changes requested):

Evgeniy-L (commented):
Looks ok, please consider adding a test for "fail fast"

Evgeniy-L (approved):
LGTM

PR #284, opened 2021-08-26
Remove uses of protoc-gen-go/generator

shani1998:
This package was intended to use for internal implementation of `protoc-gen-go` https://github.com/golang/protobuf/issues/1104#issuecomment-619420855
```
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.
```
Above warning causing problem for onprem/cloud components to format the logs into JSON.  and making hard to use `jq`

### DIFF
--- a/auth/context.go
+++ b/auth/context.go
@@ -1,10 +1,10 @@
-package util
+package auth
 
 import (
 	"context"
+
 	"google.golang.org/grpc/metadata"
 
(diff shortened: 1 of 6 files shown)


drewwells (approved):
I dig it, less external dependencies is a good thing

PR #288, opened 2021-09-15
Fix concurrent use of UnaryServerInterceptor

AnPiakhota-Infoblox:
The PR fixes the issue when multiple requests are interacting with database at the same time and the failure occurs due to the common instance of transaction linked to the interceptor function from outside its body. 

```go
    func UnaryServerInterceptor(db *gorm.DB) grpc.UnaryServerInterceptor {
    	txn := &Transaction{parent: db}
    	return UnaryServerInterceptorTxn(txn)
    }

    func UnaryServerInterceptorTxn(txn *Transaction) grpc.UnaryServerInterceptor {
    	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
    	    // here the same txn instance is being processed on every request	
    	    defer func() {
            ...
```

The interceptor is a closure since PR [#243](https://github.com/infobloxopen/atlas-app-toolkit/pull/243) 

Tested with unit tests and by replacing atlas-app-toolkit in atlas.tagging where the issue emerged after upgrading atlas-app-toolkit  to the latest version (v1.1.0)

### DIFF
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -47,6 +47,7 @@ type Transaction struct {
 	parent          *gorm.DB
 	current         *gorm.DB
 	afterCommitHook []func(context.Context)
+	configFn        []func() *Transaction // configFn holds a set of functions to configure the transaction state if neccessary.
 }
 
 func NewTransaction(db *gorm.DB) Transaction {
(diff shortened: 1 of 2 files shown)


Evgeniy-L (approved):

Evgeniy-L:
Should fix a regression caused by #243

PR #297, opened 2021-11-19
Tracing fixes for error span

mphillips-infoblox:
Fixes for adding error spans caused by bug in open telemetry collector.

See discussion here: https://github.com/open-telemetry/opentelemetry-collector/pull/70
* `code` should be added as a `census.status_code` tag.
* `message` should be added as a `census.status_description` tag.
* If code is NOT StatusCodeOK, then add `error` tag with value `true`

### DIFF
--- a/tracing/span.go
+++ b/tracing/span.go
@@ -78,16 +78,15 @@ func AddErrorCurrentSpan(ctx context.Context, err error) error {
 
 //AddErrorSpan adds error into span
 func AddErrorSpan(span *trace.Span, err error) error {
-	var code int32 = trace.StatusCodeUnknown
+	var code int32 = trace.StatusCodeOK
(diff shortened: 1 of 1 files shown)


tracing/span.go
```diff
@@ -78,16 +78,15 @@ func AddErrorCurrentSpan(ctx context.Context, err error) error {
 
 //AddErrorSpan adds error into span
 func AddErrorSpan(span *trace.Span, err error) error {
-	var code int32 = trace.StatusCodeUnknown
+	var code int32 = trace.StatusCodeOK
 	status, ok := status.FromError(err)
```
kd7lxl, line 82: The reuse of `status` here from module to variable makes the code harder to understand.

kd7lxl (approved):

PR #304, opened 2022-02-15
Upgrading to gorm v2

wk8:


calvin2021y:
Hi @wk8, Thanks for the nice work.

I try use your branch but get this error:

```sh
user.pb.gorm.go:52:41: cannot use m.Id (type *"github.com/infobloxopen/atlas-app-toolkit/atlas/resource".Identifier) as type *"github.com/infobloxopen/atlas-app-toolkit/rpc/resource".Identifier in argument to "github.com/infobloxopen/atlas-app-toolkit/gorm/resource".Decode
```

There is  atlas.rpc in your  `atlas-app-toolkit jrouge/gorm_v2` branch, but also there is `atlas.resource.v1`,  how to fix this conflict ?

calvin2021y:
I try add this patch:
```patch
diff --git a/gorm/resource/resource.go b/gorm/resource/resource.go
index 6d78620..4445134 100644
--- a/gorm/resource/resource.go
+++ b/gorm/resource/resource.go
@@ -9,7 +9,7 @@ import (
 
        "github.com/golang/protobuf/proto"
 
-       resourcepb "github.com/infobloxopen/atlas-app-toolkit/rpc/resource"
+       resourcepb "github.com/infobloxopen/atlas-app-toolkit/atlas/resource"
        "github.com/infobloxopen/atlas-app-toolkit/util"
 )

```

then get this error:
```sh
# github.com/infobloxopen/atlas-app-toolkit/gorm/resource
atlas-app-toolkit/gorm/resource/resource.go:124:5: undefined: resource.Nil
atlas-app-toolkit/gorm/resource/resource.go:130:10: undefined: resource.BuildString
atlas-app-toolkit/gorm/resource/resource.go:240:56: undefined: resource.ParseString
```

gorm/utilities.go
```diff
@@ -214,3 +213,9 @@ type EmptyFieldPathError struct {
 func (e *EmptyFieldPathError) Error() string {
 	return fmt.Sprintf("Empty field path is not allowed")
 }
+
+var namingStrategy = schema.NamingStrategy{}
```
Calebjh, line 217: Is this required? It looks unused (supposed to be used in L220?), but I know sometimes these statements do more than they seem.

gorm/transaction.go
```diff
@@ -139,25 +133,42 @@ func (t *Transaction) Rollback() error {
...
 // Reset current transaction and returns an error if any.
 func (t *Transaction) Commit(ctx context.Context) error {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
-	if t.current == nil || reflect.ValueOf(t.current.CommonDB()).IsNil() {
+	if t.current == nil {
+		return nil
+	}
+	if sqlDB, err := t.current.DB(); err == nil && reflect.ValueOf(sqlDB).IsNil() {
```
Calebjh, line 168: This seems to have a new potential `err` case that is just dropped. Probably not a problem, but it doesn't look very clean

types/jsonb.go
```diff
@@ -0,0 +1,30 @@
+package types
+
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+type Jsonb struct {
```
Calebjh, line 10: Is this a workaround to keep a simpler JSON type and not have to worry about merging the handling logic for the new JSON gorm type? https://github.com/go-gorm/datatypes/blob/master/json.go

Calebjh (commented):
Sorry for the disgustingly late PR review. If you're still interested in contributing this, I have a couple questions, but it's pretty good as-is.

types/jsonb.go
wk8: Indeed, that's better, changing

gorm/transaction.go
wk8: Fair enough, changed

Calebjh:
Merged into a new gorm_v2 branch in this repo, too, until we can get the protoc-gen-gorm and any other repos that depend on it into a state that also work with the gorm V2 changes

PR #331, opened 2022-07-06
Feature/search-by

abalaven:


gorm/collection_operators.go
```diff
@@ -62,6 +67,40 @@ func ApplyCollectionOperatorsEx(ctx context.Context, db *gorm.DB, obj interface{
 	return db, nil
 }
 
+func ApplyCollectionOperatorsWithSearchingEx(ctx context.Context, db *gorm.DB, obj interface{}, c CollectionOperatorsConverter, f *query.Filtering, s *query.Sorting, p *query.Pagination, fs *query.FieldSelection, sc *query.Searching, fieldsForFTS []string) (*gorm.DB, error) {
```
Calebjh, line 70: Because this is a new operator with a separate function, it won't be handled in the normal protoc-gen-gorm generated handlers without changes there.

Calebjh (approved):
This seems okay, and doesn't look like it has any breaking changes to be worried about. It should have added documentation in the README for the new collection operator and what it does though.

query/collection_operators.proto
```diff
@@ -221,3 +221,16 @@ message PageInfo {
...
+message Searching {
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+        json_schema: {
+            type: STRING;
+            description: "atlas.api.search_by";
+        };
+    };
+
+    string query = 1;
+}
```
Calebjh, line 236: Missing new-line at EOF

query/README.md
```diff
@@ -162,3 +164,10 @@ server.WithGateway(
   )
 )
 ```
+## Searching
+
+The syntax of REST representation of `infoblox.api.Searching` is the following.
+
+| Request Parameter | Description                              |
+| ----------------- |------------------------------------------|
+| _fts              | A string expression which performs a full-text-search on the DB |
```
Calebjh, line 173: Is "full-text-search" a commonly understood term? I think it would be good to define what exactly it means, and possibly give an example or two

PR #337, opened 2022-07-22
Case insensitive matching of field names

ychen-bloxer:
there is an issue returning the values for an object when requesting the list of DNS Server Groups (API object auth_nsg) via API when using the fields.

When there is no specific filter/tag applied in the query, everything looks good and the complete data is displayed.

### DIFF
--- a/gorm/collection_operators.go
+++ b/gorm/collection_operators.go
@@ -16,7 +16,7 @@ type SortingCriteriaConverter interface {
 }
 
 type FieldSelectionConverter interface {
-	FieldSelectionToGorm(ctx context.Context, fs *query.FieldSelection, obj interface{}, ignoreCase ...bool) ([]string, error)
+	FieldSelectionToGorm(ctx context.Context, fs *query.FieldSelection, obj interface{}) ([]string, error)
 }
 
 type PaginationConverter interface {
--- a/gorm/collection_operators.go
+++ b/gorm/collection_operators.go
@@ -16,7 +16,7 @@ type SortingCriteriaConverter interface {
 }
 
 type FieldSelectionConverter interface {
-	FieldSelectionToGorm(ctx context.Context, fs *query.FieldSelection, obj interface{}) ([]string, error)
+	FieldSelectionToGorm(ctx context.Context, fs *query.FieldSelection, obj interface{}, ignoreCase ...bool) ([]string, error)
 }
 
 type PaginationConverter interface {
--- a/gorm/fields.go
+++ b/gorm/fields.go
@@ -3,7 +3,6 @@ package gorm
 import (
 	"context"
 	"fmt"
-	"github.com/infobloxopen/atlas-app-toolkit/util"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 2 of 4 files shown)


chinmayb:
Please give some more details of the fix and please avoid using internal tickets.

gorm/fields.go
```diff
@@ -76,35 +76,53 @@ fields:
...
+		for i := 0; i < objType.NumField(); i++ {
+			if strings.EqualFold(name, strings.ToLower(strings.ReplaceAll(queryFieldName, "_", ""))) {
+				return true
+			}
+		}
+		return false
+	})
+
 	if !ok {
+		fmt.Printf("===> no field found")
```
daniel-garcia, line 92: Use log package

daniel-garcia (changes requested):

ychen-bloxer:
link issue https://infoblox.atlassian.net/browse/ATLAS-13969

gorm/fields.go
```diff
@@ -3,37 +3,39 @@ package gorm
 import (
 	"context"
 	"fmt"
+	"github.com/infobloxopen/atlas-app-toolkit/util"
```
brocwoodworthIBLX, line 6: please revert the change in import order. usually you can run `go fmt ./...` in the parent directory and it will sort these for you

gorm/fields.go
```diff
@@ -75,36 +77,55 @@ fields:
 	return toPreload, nil
 }
 
-func handlePreloads(f *query.Field, objType reflect.Type) ([]string, error) {
-	sf, ok := objType.FieldByName(util.Camel(f.GetName()))
+func handlePreloads(f *query.Field, objType reflect.Type, ignoreCase ...bool) ([]string, error) {
+	queryFieldName := f.GetName()
+	fmt.Printf("Query name = %v\n", queryFieldName)
```
brocwoodworthIBLX, line 82: use log package here instead of fmt

gorm/fields.go
```diff
@@ -75,36 +77,55 @@ fields:
...
+
+	var sf reflect.StructField
+	var ok bool
+	if len(ignoreCase) > 0 && ignoreCase[0] {
```
brocwoodworthIBLX, line 86: maybe I'm missing something but why is `ignoreCase` a `...bool` instead of just a plain old `bool`?

brocwoodworthIBLX (commented):
couple of minor things and a question to be addressed

gorm/fields.go
ychen-bloxer: This is an "optional" parameter for backward compatibility.

brocwoodworthIBLX: ok that makes sense. if that's the case then make sure that there is a unti test for `ignoreCase` being `false`, and one for there being no `ignoreCase` value at all

ychen-bloxer: Yes, there are. In test file, the function has been called as followed
	FieldSelectionStringToGorm(context.Background(), test.fs, &Model{}) // no "ignoreCase" = false
	FieldSelectionStringToGorm(context.Background(), test.fs, &Model{}, true)

brocwoodworthIBLX: it might be nice to have at least one `FieldSelectionStringToGorm(context.Background(), test.fs, &Model{}, false)` just for completion's sake

brocwoodworthIBLX (approved):
LGTM

gorm/collection_operators.go
```diff
@@ -16,7 +16,7 @@ type SortingCriteriaConverter interface {
 }
 
 type FieldSelectionConverter interface {
-	FieldSelectionToGorm(ctx context.Context, fs *query.FieldSelection, obj interface{}) ([]string, error)
+	FieldSelectionToGorm(ctx context.Context, fs *query.FieldSelection, obj interface{}, ignoreCase ...bool) ([]string, error)
```
daniel-garcia, line 19: this needs to be a option struct{} not a boolean. consider the meaning is for multiple options.

daniel-garcia (changes requested):

daniel-garcia (approved):
LGTM.

PR #339, opened 2022-08-02
Render numbers read from JWTs as integers

Calebjh:
Currently, if there is a long JWT field, e.g. `"ID":"123456789"`, this function will parse and return it as `1.23456789E8`, because all JWT numbers are treated as float64s. This could still be wonky if there are _really_ big numbers in the JWT field, since they'll now render with lots of zeros, but this is probably better for the majority of cases.


markblanford (approved):
✅

brocwoodworthIBLX (approved):

PR #349, opened 2022-11-23
Better default response codes from HTTP method when no GRPC code set

Calebjh:


gateway/status_test.go
```diff
@@ -29,14 +29,48 @@ func TestStatus(t *testing.T) {
...
+	}
+
+	stat, statName = HTTPStatus(context.Background(), "DELETE", nil)
+	if stat != http.StatusNoContent {
+		t.Errorf("invalid http status code %d - expected: %d", stat, http.StatusNoContent)
+	}
+	if statName != CodeName(Deleted) {
+		t.Errorf("invalid http status codename %q - expected: %q", statName, CodeName(Deleted))
+	}
 }
```
jcookson-infoblox, line 74: what about "PATCH"

jcookson-infoblox: tanks for adding it.

daniel-garcia (approved):

PR #363, opened 2023-01-28
Refactoring: update go version and go modules, and remove cogs

hramazani:


brocwoodworthIBLX:
is there a specific reason for removing cogs?

hramazani:
> is there a specific reason for removing cogs?

It is moved to another repository.

Calebjh (approved):
probably okay

PR #413, opened 2024-02-16
feature/compartment

sbhagate-infoblox:


auth/jwt.go
```diff
@@ -32,6 +35,12 @@ var (
 		MultiTenancyField,
 		"AccountID",
 	}
+
+	// multiCompartmentVariants all possible multi-compartment names
+	multiCompartmentVariants = []string{
+		MultiCompartmentField,
+		"CompartmentID",
+	}
```
drewwells, line 43:
```suggestion
```

Let's not do this. You have defined the field name as `compartment_id`, there's no reason for it to then attempt to read `CompartmentID`. I have no idea why AccountID supports two variations, probably backwards compatibility to dead code

auth/jwt.go
```diff
@@ -74,6 +83,22 @@ func GetAccountID(ctx context.Context, keyfunc jwt.Keyfunc) (string, error) {
...
+	for _, compartmentField := range multiCompartmentVariants {
+		val, err = GetJWTField(ctx, compartmentField, keyfunc)
+		if err == nil {
+			return val, nil
+		}
+	}
+	if err == errMissingField {
+		return val, nil
+	}
+	return val, err
```
drewwells, line 99:
```suggestion
	return GetJWTField(ctx, compartmentField, keyfunc)
```

drewwells (changes requested):
I like that the key is a string. Consumers can write their own getters. Let's get rid of the multivarient thing that account_id uses. The context key is `var string = "compartment_id"` no reason to support any other key.

auth/jwt.go
```diff
@@ -74,6 +77,18 @@ func GetAccountID(ctx context.Context, keyfunc jwt.Keyfunc) (string, error) {
...
+// GetCompartmentID gets the JWT from a context and returns the CompartmentID field
+func GetCompartmentID(ctx context.Context, keyfunc jwt.Keyfunc) (string, error) {
+	var val string
+	var err error
+	val, err = GetJWTField(ctx, MultiCompartmentField, keyfunc)
+	if err == nil || err == errMissingField {
+		return val, nil
+	} else {
+		return val, err
+	}
```
drewwells, line 89: This does the same thing. Guess I'm fine with ignoring missing compartment errors.
```suggestion
	val, err := GetJWTField(ctx, MultiCompartmentField, keyfunc)
	if err == errMissingField {
		return "", nil
	}
	return val, err
```

PR #426, opened 2024-11-13
DDINGI-1306: Revised Data Paging/Totals

Ramky-Infoblox:
DDINGI-1306: Revised Data Paging/Totals

Enhancement to the existing pagination with resultset count with enable/disable count feature.


query/collection_operators.proto
```diff
@@ -202,6 +202,9 @@ message Pagination {
     // The service may impose maximum value.
     // If omitted the service may impose a default value.
     int32 limit = 3;
+    //The bool value enables/disables the record count.
+    //The default setting disables the record count.
+    bool is_total_size = 4;
```
chinmayb, line 207: How about this ? cc @pnagaraj 
```suggestion
    // The bool value enables/disables the record count.
    // The default setting disables the record count.
    bool is_total_count_required = 4;
```

query/collection_operators.proto
```diff
@@ -220,6 +223,8 @@ message PageInfo {
     // The service may optionally include the offset of the next page of resources.
     // A null value indicates no more pages.
     int32 offset = 3;
+    //Resultset record count.
+    int64 total_size = 4;
```
chinmayb, line 227:
```suggestion
    // total_size indicates the total records present.
    int64 total_size = 4;
```

query/collection_operators.proto
Ramky-Infoblox: @chinmayb - This is not a required field. 
As comment says it is just to enable or disable the record count.

query/collection_operators.proto
Ramky-Infoblox: sure will modify

query/collection_operators.proto
Ramky-Infoblox: for comment space will modify it

chinmayb: the user asks this, hence I was asking that way. It doesn't indicate that its a required field or anything. More like
```
bool is_total_size_needed
```

Ramky-Infoblox: Sure

chinmayb (approved):

PR #429, opened 2024-12-18
promoting changes to add support for read only db instance

aankam-infoblox:
This is to support read only database instance for performing read  operations on the database. Once these changes are approved and merged then all the modules can use this functionality to implement read only database support to perform database read operations in their project.

### DIFF
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -109,19 +109,17 @@ func getReadOnlyDBTxn(ctx context.Context, opts *databaseOptions, txn *Transacti
 	case txn.parentRO == nil:
 		return getReadWriteDBTxn(ctx, opts, txn)
 	case opts.txOpts != nil && txn.currentOpts.txOpts != nil:
-		if opts.txOpts.ReadOnly != txn.currentOpts.txOpts.ReadOnly || opts.txOpts.Isolation != txn.currentOpts.txOpts.Isolation {
+		if *opts.txOpts != *txn.currentOpts.txOpts {
 			return nil, ErrCtxTxnOptMismatch
 		}
-	default:
+	case opts.txOpts != nil:
 		// We should error in two cases 1. We should error if read-only DB requested with read-write txn
 		// 2. If no txn options provided in previous call but provided in subsequent call
-		if opts.txOpts != nil {
-			if opts.txOpts.ReadOnly == false || txn.currentOpts.database != dbNotSet {
-				return nil, ErrCtxTxnOptMismatch
-			}
-			txnOpts := *opts.txOpts
-			txn.currentOpts.txOpts = &txnOpts
+		if opts.txOpts.ReadOnly == false || txn.currentOpts.database != dbNotSet {
+			return nil, ErrCtxTxnOptMismatch
 		}
+		txnOpts := *opts.txOpts
+		txn.currentOpts.txOpts = &txnOpts
 	}
 	if txn.current != nil {
 		return txn.current, nil
@@ -143,18 +141,16 @@ func getReadWriteDBTxn(ctx context.Context, opts *databaseOptions, txn *Transact
 	case txn.parent == nil:
 		return nil, ErrCtxTxnNoDB
 	case opts.txOpts != nil && txn.currentOpts.txOpts != nil:
-		if opts.txOpts.ReadOnly != txn.currentOpts.txOpts.ReadOnly || opts.txOpts.Isolation != txn.currentOpts.txOpts.Isolation {
+		if *opts.txOpts != *txn.currentOpts.txOpts {
 			return nil, ErrCtxTxnOptMismatch
 		}
-	default:
-		if opts.txOpts != nil {
-			// We should return error If no txn options provided in previous call but provided in subsequent call
-			if txn.currentOpts.database != dbNotSet {
-				return nil, ErrCtxTxnOptMismatch
-			}
-			txnOpts := *opts.txOpts
-			txn.currentOpts.txOpts = &txnOpts
+	case opts.txOpts != nil:
+		// We should return error If no txn options provided in previous call but provided in subsequent call
+		if txn.currentOpts.database != dbNotSet {
+			return nil, ErrCtxTxnOptMismatch
 		}
+		txnOpts := *opts.txOpts
+		txn.currentOpts.txOpts = &txnOpts
 	}
 	if txn.current != nil {
 		return txn.current, nil
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -72,7 +72,7 @@ type DatabaseOption func(*databaseOptions)
 // WithRODB returns closure to set the readOnlyReplica flag
 func WithRODB(readOnlyReplica bool) DatabaseOption {
 	return func(ops *databaseOptions) {
-		if readOnlyReplica == false {
+		if !readOnlyReplica {
 			ops.database = dbReadWrite
 		} else {
 			ops.database = dbReadOnly
@@ -115,7 +115,7 @@ func getReadOnlyDBTxn(ctx context.Context, opts *databaseOptions, txn *Transacti
 	case opts.txOpts != nil:
 		// We should error in two cases 1. We should error if read-only DB requested with read-write txn
 		// 2. If no txn options provided in previous call but provided in subsequent call
-		if opts.txOpts.ReadOnly == false || txn.currentOpts.database != dbNotSet {
+		if !opts.txOpts.ReadOnly || txn.currentOpts.database != dbNotSet {
 			return nil, ErrCtxTxnOptMismatch
 		}
 		txnOpts := *opts.txOpts
--- a/gorm/transaction.go
+++ b/gorm/transaction.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/infobloxopen/atlas-app-toolkit/rpc/errdetails"
 	"github.com/jinzhu/gorm"
-	logger "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -19,24 +18,25 @@ import (
 // This prevents collisions with keys defined in other packages.
 type ctxKey int
 
-// readOnlyDBKey is an unexported type and used to define a key for storing read only db instance in the context.
-// This prevents collisions with keys defined in other package
-type readOnlyDBKey int
+// This is used to define various options to set/get the current database setup in the context
+type dbType int
 
 // txnKey is the key for `*Transaction` values in `context.Context`.
 // It is unexported; clients use NewContext and FromContext
 // instead of using this key directly.
 var txnKey ctxKey
 
-// roDBKey is the key used for storing read-only db instance in the context.
-// It is unexported; clients use BeginFromContext with options to get the read only db instance
-// instead of using this key directly.
-var roDBKey readOnlyDBKey
-
 var (
-	ErrCtxTxnMissing = errors.New("Database transaction for request missing in context")
-	ErrCtxTxnNoDB    = errors.New("Transaction in context, but DB is nil")
-	ErrNoReadOnlyDB  = errors.New("No read-only DB")
+	ErrCtxTxnMissing     = errors.New("Database transaction for request missing in context")
+	ErrCtxTxnNoDB        = errors.New("Transaction in context, but DB is nil")
+	ErrCtxTxnOptMismatch = errors.New("Transaction in context, but Txn opts are mismatched")
+	ErrCtxDBOptMismatch  = errors.New("Transaction in context, but DB opts are mismatched")
+)
+
+const (
+	dbNotSet dbType = iota
+	dbReadOnly
+	dbReadWrite
 )
 
 // NewContext returns a new Context that carries value txn.
@@ -56,24 +56,33 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
 type Transaction struct {
 	mu              sync.Mutex
 	parent          *gorm.DB
+	parentRO        *gorm.DB
 	current         *gorm.DB
-	readOnly        bool
+	txOpts          *sql.TxOptions
+	currentDB       dbType
 	afterCommitHook []func(context.Context)
 }
 
 type databaseOptions struct {
 	readOnlyReplica bool
+	txOpts          *sql.TxOptions
 }
 
 type DatabaseOption func(*databaseOptions)
 
-// WithRODB returns clouser to set the readOnlyReplica flag
+// WithRODB returns closure to set the readOnlyReplica flag
 func WithRODB(readOnlyReplica bool) DatabaseOption {
 	return func(ops *databaseOptions) {
 		ops.readOnlyReplica = readOnlyReplica
 	}
 }
 
+// WithTxOptions returns a closure to set the TxOptions
+func WithTxOptions(opts *sql.TxOptions) DatabaseOption {
+	return func(ops *databaseOptions) {
+		ops.txOpts = opts
+	}
+}
 func toDatabaseOptions(options ...DatabaseOption) *databaseOptions {
 	opts := &databaseOptions{}
 	for _, op := range options {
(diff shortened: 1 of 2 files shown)


gorm/transaction.go
```diff
@@ -39,13 +43,33 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
 	return
 }
 
+// GetReadOnlyDB returns the read only db instance stored in the ctx if there is no txn in use with read/write database
+func GetReadOnlyDB(ctx context.Context) (*gorm.DB, error) {
```
davidhankins, line 47: Let's add unit tests for this new function, see `transaction_test.go`

gorm/transaction.go
```diff
@@ -57,15 +81,29 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
 // As result new instance of `*gorm.DB` will be returned.
 // Error will be returned in case either transaction or db connection info is missing in context.
 // Gorm specific error can be checked by `*gorm.DB.Error`.
 func BeginFromContext(ctx context.Context) (*gorm.DB, error) {
+	logger := ctxlogrus.Extract(ctx)
 	txn, ok := FromContext(ctx)
 	if !ok {
 		return nil, ErrCtxTxnMissing
 	}
+	if txn.readOnly == true {
```
davidhankins, line 103: We'll need to update the `BeginFromContext` tests for this change. The existing tests do not check that a transaction from the read-write parent is returned in the current cases, and we'll have to add a test to confirm the read-only database reference is returned on a read-only context.

gorm/transaction.go
```diff
@@ -220,6 +258,12 @@ func UnaryServerInterceptorTxn(txn *Transaction) grpc.UnaryServerInterceptor {
 		}()
 
 		ctx = NewContext(ctx, txn)
+		if len(readOnlyDB) > 0 {
```
davidhankins, line 261: Let's add a check for this to `TestUnaryServerInterceptorTxn_success`

gorm/transaction.go
```diff
@@ -57,15 +81,29 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
 	t.afterCommitHook = append(t.afterCommitHook, hooks...)
 }
 
-// BeginFromContext will extract transaction wrapper from context and start new transaction.
+// GetReadOnlyDBInstance returns the read only database instance stored in the ctx
+func GetReadOnlyDBInstance(ctx context.Context) (*gorm.DB, error) {
```
davidhankins, line 85: Let's not export this unless there's a reason to? `getReadOnlyDBInstance`

With `GetReadOnlyDB` and `BeginFromContext` tested, I don't think there's much value in specifically testing this function, so I don't think a unit test for it alone is required.

davidhankins (commented):
Yeah, I agree with the semantic changes here although they appear odd.

To try and explain my reasoning;

I think the `Transaction struct` is ~poorly named - it can produce one transaction from a parent database in the current ctx - and deliver that same transaction to all callers in that ctx's call graph. This is more of a "context database state" structure than specifically a transaction. To wit, we use `BeginFromContext(ctx)` regularly in read-only functions deeply nested in the call graph of API calls to avoid "plumbing" (passing the transaction reference as an argument down a long call chain).

So it's fine that a caller asking for a transaction is instead given a read-only-database reference because that's actually what they're after ("the database reference for this api request, read or write").

Probably on the master branch we might want to think about renaming or providing a new interface to the database-context that tries to clear up some of the confusing terminology.

Just some minor nits on my part - let's get some unit tests drafted and see about merging this.

gorm/transaction.go
```diff
@@ -17,15 +18,18 @@ import (
 // ctxKey is an unexported type for keys defined in this package.
 // This prevents collisions with keys defined in other packages.
 type ctxKey int
+type readOnlyDBKey int
```
akleinib, line 21: We have comments for existing fields, we should keep this pattern when adding new ones.

gorm/transaction.go
```diff
@@ -39,13 +43,33 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
...
 
+// GetReadOnlyDB returns the read only db instance stored in the ctx if there is no txn in use with read/write database
+func GetReadOnlyDB(ctx context.Context) (*gorm.DB, error) {
+	logger := ctxlogrus.Extract(ctx)
+	txn, ok := FromContext(ctx)
+	if !ok {
+		return nil, ErrCtxTxnMissing
+	}
+	if txn.current != nil {
+		logger.Warnf("GetReadOnlyDB: Txn already initialized with read/write DB")
```
akleinib, line 54: 1. Maybe the caller would like to know that it asked for a read-only transaction and got a read-write one instead.
2. Extracting the logger from the context assumes this code is used with the gRPC gateway which might not be the case.

gorm/transaction.go
```diff
@@ -57,15 +81,29 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
+// GetReadOnlyDBInstance returns the read only database instance stored in the ctx
+func GetReadOnlyDBInstance(ctx context.Context) (*gorm.DB, error) {
+	dbRO, ok := ctx.Value(roDBKey).(*gorm.DB)
+	if !ok {
+		return nil, ErrNoReadOnlyDB
+	}
+	return dbRO, nil
+}
+
+// BeginFromContext will extract transaction wrapper from context and start new transaction if transaction is not set to read only otherwise it will return read only database instance
```
akleinib, line 93: Not using an actual transaction for read-only operations introduces two problems:
1. No support for isolation level,
2. Performance (see `Statements are executed more quickly in a transaction block, because transaction start/commit requires significant CPU and disk activity.` in https://www.postgresql.org/docs/15/sql-begin.html).

akleinib:
It would be good to document here how this is supposed to be used. Do we need to add something like the following everywhere to fall back on R/W replica?

```go
txn, err := GetReadOnlyDB(ctx)

if errors.Is(err, ErrNoReadOnlyDB) {
   txn, err = BeginFromContext(ctx)
}
```

I believe it would be much better to implement this using the standard "option" pattern, e.g.:
```go
func BeginFromContext(ctx context.Context, OPTIONS...) (*gorm.DB, error) { }
```

`OPTIONS` can be:
* If none, current behavior,
* Support of https://pkg.go.dev/database/sql#TxOptions,
* From read-only replica, if it doesn't exist fall back to read/write,
* From read-only replica, if it doesn't exist return an error,
* ...

This way we can support all the current and future use cases rather keep adding functions with new prototypes.

gorm/transaction.go
```diff
@@ -18,14 +19,24 @@ import (
 // This prevents collisions with keys defined in other packages.
 type ctxKey int
 
+// readOnlyDBKey is an unexported type and used to define a key for storing read only db instance in the context.
```
akleinib, line 22: Rather than storing the read-only database in the context, cannot we store it in the `Transaction` structure like the existing database? A `nil` value for this variable indicates no read-only database.

gorm/transaction.go
```diff
@@ -46,9 +57,31 @@ type Transaction struct {
...
 	afterCommitHook []func(context.Context)
 }
 
+type databaseOptions struct {
+	readOnlyReplica bool
+}
+
+type DatabaseOption func(*databaseOptions)
+
+// WithRODB returns clouser to set the readOnlyReplica flag
```
akleinib, line 70: Typo `clouser`.

gorm/transaction.go
```diff
@@ -57,15 +90,50 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
+		if txn.current == nil {
+			return getReadOnlyDBInstance(ctx)
+		} else {
+			logger.Warnf("BeginFromContext: requested: read-only DB, returns: read-write DB, reason: read-write DB txn in use")
+			return txn.current, nil
+		}
+	} else if txn.readOnly == true {
+		logger.Warnf("BeginFromContext: requested: read-write DB, returns: read-only DB, reason: txn set to read only")
+		return getReadOnlyDBInstance(ctx)
+	}
```
akleinib, line 136: Returning a R/W transaction instead of R-O transaction is ok since all the read-only operations will be executed by the R/W transaction. Doing the inverse doesn't work. This should return an error.

gorm/transaction.go
```diff
@@ -57,15 +90,50 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
 // As result new instance of `*gorm.DB` will be returned.
 // Error will be returned in case either transaction or db connection info is missing in context.
 // Gorm specific error can be checked by `*gorm.DB.Error`.
-func BeginFromContext(ctx context.Context) (*gorm.DB, error) {
+func BeginFromContext(ctx context.Context, options ...DatabaseOption) (*gorm.DB, error) {
 	txn, ok := FromContext(ctx)
 	if !ok {
 		return nil, ErrCtxTxnMissing
 	}
+	opts := toDatabaseOptions(options...)
```
akleinib, line 125: I would keep the check of `txn.parent` before adding all this new code.

gorm/transaction.go
```diff
@@ -57,15 +90,50 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
+	if opts.readOnlyReplica == true {
+		if txn.current == nil {
+			return getReadOnlyDBInstance(ctx)
+		} else {
+			logger.Warnf("BeginFromContext: requested: read-only DB, returns: read-write DB, reason: read-write DB txn in use")
```
akleinib, line 130: It might be better to provide a `WithLogger()` function people can use to provide their logger instance. This way all these messages are logged with the same logger as the rest of the application. If no logger is provided, we can use a global one.

gorm/transaction.go
```diff
@@ -46,9 +57,31 @@ type Transaction struct {
...
 }
 
+type databaseOptions struct {
+	readOnlyReplica bool
+}
+
+type DatabaseOption func(*databaseOptions)
+
+// WithRODB returns clouser to set the readOnlyReplica flag
+func WithRODB(readOnlyReplica bool) DatabaseOption {
```
akleinib, line 71: We should add a "with function" for `sql.TxOptions` since we have existing use-cases.

gorm/transaction.go
akleinib: This point is not addressed. The code still returns the "database" and not a transaction.

gorm/transaction.go
davidhankins: I wonder if this should be at the UnaryServerInterceptor level rather than the Transaction level.

gorm/transaction.go
```diff
@@ -57,15 +90,50 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
+	opts := toDatabaseOptions(options...)
+	if opts.readOnlyReplica == true {
+		if txn.current == nil {
+			return getReadOnlyDBInstance(ctx)
+		} else {
+			logger.Warnf("BeginFromContext: requested: read-only DB, returns: read-write DB, reason: read-write DB txn in use")
+			return txn.current, nil
+		}
+	} else if txn.readOnly == true {
+		logger.Warnf("BeginFromContext: requested: read-write DB, returns: read-only DB, reason: txn set to read only")
```
davidhankins, line 134: I think this is normal now, and the log line can be removed.

gorm/transaction.go
```diff
@@ -45,10 +56,41 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
 type Transaction struct {
 	mu              sync.Mutex
 	parent          *gorm.DB
+	parentRO        *gorm.DB
 	current         *gorm.DB
+	txOpts          *sql.TxOptions
```
davidhankins, line 61: Consider using e.g. `currentOpts *databaseOptions` rather than separately `txOpts` and `currentDB` etc.

I think you could more directly compare "current and prior options" that way; just keep the options value from when the transaction was started and compare options to options.

gorm/transaction.go
```diff
@@ -45,10 +56,41 @@ func FromContext(ctx context.Context) (txn *Transaction, ok bool) {
...
+	currentDB       dbType
 	afterCommitHook []func(context.Context)
 }
 
+type databaseOptions struct {
+	readOnlyReplica bool
```
davidhankins, line 67: Consider using `dbType` here (so you can have notset / readonly / readwrite as a trinary instead of boolean value).

gorm/transaction.go
```diff
@@ -57,25 +99,107 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
-// Error will be returned in case either transaction or db connection info is missing in context.
-// Gorm specific error can be checked by `*gorm.DB.Error`.
-func BeginFromContext(ctx context.Context) (*gorm.DB, error) {
-	txn, ok := FromContext(ctx)
-	if !ok {
-		return nil, ErrCtxTxnMissing
+// getReadOnlyDBInstance returns the read only db txn if RO DB available otherwise it returns read/write db txn
+func getReadOnlyDBTxn(ctx context.Context, opts *databaseOptions, txn *Transaction) (*gorm.DB, error) {
+	var db *gorm.DB
+	if txn.parentRO == nil {
```
davidhankins, line 105: Try to restructure the if / else / if / if / else / ... chain here as a `switch {`.

You might have to have nested levels of `switch {` or some `if` conditions inside cases, but I think it would read better e.g.

```
switch {
  case txn.parentRO == nil:
    return ...
  case opts.txOpts != nil:
    ... handle case where txn.txOpts is or isn't nil (nested if)
  case txn.txOpts != nil:
    ... at this point you know opts.txOpts is nil.
  default:
    ...
}
```

gorm/transaction.go
```diff
@@ -57,25 +99,107 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
+		if opts.txOpts != nil || txn.txOpts != nil {
+			if opts.txOpts == nil {
+				return nil, ErrCtxTxnOptMismatch
+			} else if txn.txOpts == nil {
+				if opts.txOpts.ReadOnly == false { // For Read Only DB, the Txn should be set as read only before calling this method
+					return nil, ErrCtxTxnOptMismatch
+				}
+				txn.txOpts = &sql.TxOptions{}
+				txn.txOpts.ReadOnly = opts.txOpts.ReadOnly
+				txn.txOpts.Isolation = opts.txOpts.Isolation
```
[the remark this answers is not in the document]
davidhankins: (see above comment I think we might want to just store `opts` directly rather than have two fields)

gorm/transaction_test.go
```diff
@@ -45,6 +53,95 @@ func TestUnaryServerInterceptor_success(t *testing.T) {
 	}
 }
 
+func TestUnaryServerInterceptor_with_readonlydb(t *testing.T) {
```
akleinib, line 56: Actually wondering if we need this test and the next one. They don't call `BeginFromContext()` where all the changes have been made.

gorm/transaction_test.go
```diff
@@ -329,6 +426,415 @@ func TestContext(t *testing.T) {
...
+				txn1, err := beginFromContextWithOptions(ctx, test.withOpts, test.txOpts)
+				if err != nil {
+					t.Error("Received an error beginning transaction")
+				}
+				if txn1 == nil {
+					t.Error("Did not receive a transaction from context")
+				}
+				test.withOpts = readOnly
+				_, err = beginFromContextWithOptions(ctx, test.withOpts, test.txOpts)
+				if err == nil {
```
akleinib, line 501: nit: We could check that the returned error is the expected one.

gorm/transaction_test.go
```diff
@@ -329,6 +426,415 @@ func TestContext(t *testing.T) {
...
+			readOnlyDB, dbROMock, err := sqlmock.New()
+			if err != nil {
+				t.Fatalf("failed to create sqlmock for read-only db - %s", err)
+			}
+			dbRO, err := gorm.Open("postgres", readOnlyDB)
+			if err != nil {
+				t.Fatalf("failed to open gorm read-only db - %s", err)
+			}
+			mock.ExpectBegin()
+			dbROMock.ExpectBegin()
```
akleinib, line 488: Do we really need to set expectations if we don't validate them?

gorm/transaction_test.go
```diff
@@ -329,6 +426,415 @@ func TestContext(t *testing.T) {
...
+				_, err = beginFromContextWithOptions(ctx, test.withOpts, test.txOpts)
+				if err == nil {
+					t.Error("begin transaction should fail with an error TxOptionsMismatch")
+				}
+				test.txOpts.ReadOnly = false
+				test.txOpts.Isolation = sql.LevelSerializable
+				_, err = beginFromContextWithOptions(ctx, test.withOpts, test.txOpts)
+				if err == nil {
+					t.Error("begin transaction should fail with an error TxOptionsMismatch")
+				}
```
akleinib, line 568: How this is different from the block starting on line 534?

gorm/transaction_test.go
```diff
@@ -329,6 +426,415 @@ func TestContext(t *testing.T) {
...
+				txn4, err := beginFromContextWithOptions(ctx, test.withOpts, nil)
+				if err != nil {
+					t.Error("Received an error beginning transaction")
+				}
+				if txn4 == nil {
+					t.Error("Did not receive a transaction from context")
+				}
```
akleinib, line 575: Shouldn't we check that `txn4` is the same as `txn1`?

gorm/transaction_test.go
```diff
@@ -329,6 +426,415 @@ func TestContext(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestBeginFromContextStartWithReadOnlyOptions(t *testing.T) {
```
akleinib, line 581: Same comments as in the previous case.

gorm/transaction.go
```diff
@@ -57,23 +102,102 @@ func (t *Transaction) AddAfterCommitHook(hooks ...func(context.Context)) {
...
+		// 2. If no txn options provided in previous call but provided in subsequent call
+		if opts.txOpts != nil {
+			if opts.txOpts.ReadOnly == false || txn.currentOpts.database != dbNotSet {
+				return nil, ErrCtxTxnOptMismatch
+			}
+			txnOpts := *opts.txOpts
+			txn.currentOpts.txOpts = &txnOpts
+		}
+	}
+	if txn.current != nil {
```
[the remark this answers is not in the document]
davidhankins: update: you can't move the `txn.current != nil` test into the switch, sorry about that.

gorm/transaction_test.go
```diff
@@ -329,6 +426,440 @@ func TestContext(t *testing.T) {
...
+				}
+				txn4, err := beginFromContextWithOptions(ctx, test.withOpts, nil)
+				if err != nil {
+					t.Error("Received an error beginning transaction")
+				}
+				if txn4 == nil {
+					t.Error("Did not receive a transaction from context")
+				}
+				// Case: Transaction begin is idempotent
+				if txn1 != txn4 {
```
davidhankins, line 582: i think we previous checked that `txn1 != nil` so you could omit the `txn4 == nil` case if you wanted to, but it works either way and you might want the more precise error message for the nil case, so not blocking.

davidhankins (approved):
LGTM, but we'll wait and see if Arnaud is available to confirm his questions are addressed.

akleinib (approved):

PR #432, opened 2025-01-14
return-txn-current-if-exists

brucecurcio:


akleinib:
Are these changes still prevent scenarios like:
1. Caller requested a R-O transaction first then requests a R/W transaction,
2. Caller requested a R-O transaction with isolation level A first then requests a R-O transaction with isolation level B,
3. ...

?

For those cases we shouldn't return the current transaction if the newly requested transaction doesn't match the characteristics of the existing one.

brucecurcio:
> Are these changes still prevent scenarios like:
> 
> 1. Caller requested a R-O transaction first then requests a R/W transaction,
> 2. Caller requested a R-O transaction with isolation level A first then requests a R-O transaction with isolation level B,
> 3. ...
> 
> ?
> 
> For those cases we shouldn't return the current transaction if the newly requested transaction doesn't match the characteristics of the existing one.

This is trying to address the case where a RW method has a RO method within it.  The existing atlas-app-toolkit implementation throws an error because the previous txn was RW and the new txn is RO.  It seems correct to not allow the switching between DB's (RW to RO) within the same parent method.  But it also seems as though we should not prevent the parent method from calling the child method by throwing an error.  This approach uses the previous txn if it exists.   If one wants to switch the isolation level between requests, they would have to initiate a brand new txn.  This is the case for the current implementation and for the changes in this PR.


> Are these changes still prevent scenarios like:
> 
> 1. Caller requested a R-O transaction first then requests a R/W transaction,
> 2. Caller requested a R-O transaction with isolation level A first then requests a R-O transaction with isolation level B,
> 3. ...
> 
> ?
> 
> For those cases we shouldn't return the current transaction if the newly requested transaction doesn't match the characteristics of the existing one.

as discussed, i put in a check that prevents attempts to change from RO DB to RW DB but returns the existing db txn in all other cases.

akleinib (approved):

PR #507, opened 2026-05-29
fix: reject non-string JSON types in Identifier.UnmarshalJSONPB

adychko:
Identifier.UnmarshalJSONPB previously accepted any JSON type (arrays, numbers, objects, booleans) without returning an error, producing garbage field values. This caused downstream UUID parsing failures that surfaced as 500 Internal Server Errors instead of 400 Bad Request.

Now validates that input is a JSON string or null before parsing. Returns a clear error for invalid types.

### DIFF
--- a/rpc/resource/jsonpb.go
+++ b/rpc/resource/jsonpb.go
@@ -2,6 +2,7 @@ package resource
 
 import (
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


adychko:
@Calebjh please review

